PR #166, opened 2015-01-10
Build Tornado speedups extension

### LINKED ISSUE #135: Slow web socket performance
The rosbridge web socket server is very CPU hungry when a lot of messages are being routed to browser clients.

Our use case is subscription to multiple sensors from multiple browser clients for distributed, interactive visuals.  ROS is a great platform for this kind of thing, but when we ran rosbridge on a modest server (some kind of Atom processor with plenty of RAM), it couldn't keep up.  The server would eat a lot of CPU as message traffic ramped up, and sometimes choked up resulting in apparent system breakdown.  We ended up offloading the rosbridge server to a more powerful server, but we'd like for rosbridge to be efficient enough to handle hundreds of messages per second without burning up CPU cycles on a user-facing machine.

I looked around and found this discussion pointing at Tornado's web socket masking function as the culprit, with native speedups making an order of magnitude difference:

https://groups.google.com/forum/#!topic/python-tornado/n6VpUr8zk2c

Then this Tornado pull request to add native speedups:

https://github.com/tornadoweb/tornado/pull/579

Looks like it was released in Tornado 3.2.  Ubuntu's python-tornado package is still at 3.1.1 in trusty, but pip will install 4.0.2 as of today.  Presently rosbridge includes Tornado 2.3.

This is related to #121 which should resolve this issue if Tornado comes from pip or is updated to 3.2+ in the rosbridge tree.

Thanks
mv

mvollrath:
Fail-over extension building code copied from Tornado's setup.py.

Fixes #135

### DIFF
--- a/rosbridge_tools/setup.py
+++ b/rosbridge_tools/setup.py
@@ -1,11 +1,90 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+import os
+import platform
+import sys
+import warnings
+from distutils.core import setup, Extension
 from catkin_pkg.python_setup import generate_distutils_setup
 
 d = generate_distutils_setup(
     packages=['rosbridge_tools', 'rosbridge_tools.tornado', 'rosbridge_tools.tornado.platform', 'rosbridge_tools.backports', 'rosbridge_tools.backports.ssl_match_hostname'],
     package_dir={'': 'src'}
 )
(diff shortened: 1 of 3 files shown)


mvollrath:
Looks like this was made against the wrong branch.

PR #167, opened 2015-01-10
Build Tornado speedups

### LINKED ISSUE #135: Slow web socket performance

mvollrath:
Copy fail-over build_ext code from Tornado's setup.py.

Fixes #135

### DIFF
--- a/rosbridge_server/setup.py
+++ b/rosbridge_server/setup.py
@@ -1,11 +1,90 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+import os
+import platform
+import sys
+import warnings
+
+from distutils.core import setup, Extension
 from catkin_pkg.python_setup import generate_distutils_setup
 
 d = generate_distutils_setup(
     packages=['backports', 'backports.ssl_match_hostname', 'tornado', 'tornado.platform'],
     package_dir={'': 'src'}
 )
+
+# The following code is copied from
+# https://github.com/mongodb/mongo-python-driver/blob/master/setup.py
+# to support installing without the extension on platforms where
+# no compiler is available.
+from distutils.command.build_ext import build_ext
+
+class custom_build_ext(build_ext):
+    """Allow C extension building to fail.
+
+    The C extension speeds up websocket masking, but is not essential.
+    """
+
+    warning_message = """
+********************************************************************
+WARNING: %s could not
+be compiled. No C extensions are essential for Tornado to run,
+although they do result in significant speed improvements for
+websockets.
+%s
+
+Here are some hints for popular operating systems:
+
+If you are seeing this message on Linux you probably need to
+install GCC and/or the Python development package for your
+version of Python.
+
+Debian and Ubuntu users should issue the following command:
+
+    $ sudo apt-get install build-essential python-dev
+
+RedHat, CentOS, and Fedora users should issue the following command:
+
+    $ sudo yum install gcc python-devel
+********************************************************************
+"""
+
+    def run(self):
+        try:
+            build_ext.run(self)
+        except Exception:
+            e = sys.exc_info()[1]
+            sys.stdout.write('%s\n' % str(e))
+            warnings.warn(self.warning_message % ("Extension modules",
+                                                  "There was an issue with "
+                                                  "your platform configuration"
+                                                  " - see above."))
+
+    def build_extension(self, ext):
+        name = ext.name
+        try:
+            build_ext.build_extension(self, ext)
+        except Exception:
+            e = sys.exc_info()[1]
+            sys.stdout.write('%s\n' % str(e))
+            warnings.warn(self.warning_message % ("The %s extension "
+                                                  "module" % (name,),
+                                                  "The output above "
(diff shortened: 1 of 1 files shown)


mvollrath:
I'll be happy to copy this into rosbridge_tools when it's ready.

jihoonl:
Thanks @mvollrath. I will review the module this week and merge it.

jihoonl:
I got the following error while catkin_make install process. I have confirmed that build-essential and python-dev were installed in my pc. Can you check what it is?

```
copying src/tornado/platform/kqueue.py -> /home/jihoonl/research/ros/isolate/build/rosbridge_suite/rosbridge_server/lib.linux-x86_64-2.7/tornado/platform
running build_ext
building 'tornado.speedups' extension
creating /home/jihoonl/research/ros/isolate/build/rosbridge_suite/rosbridge_server/temp.linux-x86_64-2.7
creating /home/jihoonl/research/ros/isolate/build/rosbridge_suite/rosbridge_server/temp.linux-x86_64-2.7/tornado
x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/include/python2.7 -c tornado/speedups.c -o /home/jihoonl/research/ros/isolate/build/rosbridge_suite/rosbridge_server/temp.linux-x86_64-2.7/tornado/speedups.o
x86_64-linux-gnu-gcc: error: tornado/speedups.c: No such file or directory
x86_64-linux-gnu-gcc: fatal error: no input files
compilation terminated.
/home/jihoonl/research/ros/isolate/src/rosbridge_suite/rosbridge_server/setup.py:72: UserWarning: 
********************************************************************
WARNING: The tornado.speedups extension module could not
be compiled. No C extensions are essential for Tornado to run,
although they do result in significant speed improvements for
websockets.
The output above this warning shows how the compilation failed.

Here are some hints for popular operating systems:

If you are seeing this message on Linux you probably need to
install GCC and/or the Python development package for your
version of Python.

Debian and Ubuntu users should issue the following command:

    $ sudo apt-get install build-essential python-dev

RedHat, CentOS, and Fedora users should issue the following command:

    $ sudo yum install gcc python-devel
********************************************************************

  "The output above "
command 'x86_64-linux-gnu-gcc' failed with exit status 4
running install
running install_lib
running install_egg_info
```

mvollrath:
Thanks, working on it now.

mvollrath:
Via 16445188a0f9f1fc8aeaac2576bbd43a5a05f395 I can now run catkin_make install without this warning.

mvollrath:
To verify import I've prepended install/lib/python2.7/dist-packages to PYTHONPATH and imported tornado.speedups without exception.  Anything else I should be checking?

mvollrath:
Advertise, subscribe and publish over ws all working for me.

jihoonl:
:+1: Thanks @mvollrath. It compiles and works without any error for me too. :)

mvollrath:
@jihoonl I am just now realizing that this change makes the packages specific to the build farm's arch.  At best, trying to use rosbridge_server after this patch in non-64bit env will fail to import the speedups silently (assuming it raises ImportError).  At worst, it will fail to launch.

There are probably some ways to work around this, but being able to use the system Tornado via those rosbridge_tools changes would suffice for my performance needs and be a better global solution than this PR imo.

rctoris:
@mvollrath won't the build farms arch match the arch of the client downloading it always? This should be the expected behavior from pre-built binaries.

jihoonl:
The build farm succeeded to build the new version of rosbridge_server.

http://www.ros.org/debbuild/indigo.html?q=rosbridge

mvollrath:
Cool, I just assumed the debs are architecture: any/all for some reason.  Carry on :+1:

PR #174, opened 2015-02-24
Make get_topics() and get_topic_type() reference the full list of active topics

### LINKED ISSUE #173: rosapi get_topics() should return all topics, not just published ones
Currently, rosapi's get_topics() returns published topics, but not topics marked only as being subscribed to. See https://github.com/RobotWebTools/rosbridge_suite/blob/develop/rosapi/src/rosapi/proxy.py#L50. This is probably not the expected behavior.

For example, just starting turtlebot_gazebo in Hydro, there are 39 published topics, and 14 subscribed topics, and a total of 47 unique topics. But only the 39 published topics are found by calling `rosservice call rosapi/topics`, whereas `rostopic list` shows all 47.

I think get_topics() should be implemented using Master.getSystemState() instead, as in https://github.com/RobotWebTools/rosbridge_suite/blob/develop/rosapi/src/rosapi/proxy.py#L93

jstnhuang:
Before, get_topics() only returned published topics and get_topic_type() only returned the topic type for published topics. Resolves #173.

### DIFF
--- a/rosapi/src/rosapi/proxy.py
+++ b/rosapi/src/rosapi/proxy.py
@@ -31,13 +31,13 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-from rospy import get_published_topics
 from rosservice import get_service_list
 from rosservice import get_service_type as rosservice_get_service_type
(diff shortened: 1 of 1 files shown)


jihoonl:
Thanks @jstnhuang . I will take a look and merge tomorrow.

jihoonl:
:+1: Thanks.

PR #178, opened 2015-03-31
Fixed namespacing services

reinzor:
Pull request suggestion for issue https://github.com/RobotWebTools/roslibjs/issues/173

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/services.py
+++ b/rosbridge_library/src/rosbridge_library/internal/services.py
@@ -31,7 +31,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from threading import Thread
-from rospy import ServiceProxy, resolve_name, get_namespace
+from rospy import ServiceProxy, resolve_name
 from rosservice import get_service_type
 from rosbridge_library.internal.ros_loader import get_service_class
 from rosbridge_library.internal.ros_loader import get_service_request_instance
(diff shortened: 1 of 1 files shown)


rctoris:
Looks like some tests are failing. 

If the service name is absolute (e.g., `/my_service`) I don't think anything needs to be appended otherwise you will get `//my_service`.

reinzor:
Added fix for this in commit 26493ac72a2089f5747f554cf1a888733df8625d

rctoris:
@jihoonl Does this look reasonable?

jihoonl:
It seems to support to call service in rosbridge local namespace. Would it work if for absolute namespaced service?

I think it should check leading slash of `service`  argument, then add namespace only if there is no leading slash. 

For example,  rosbridge is running under `my_namespace`. and user wants to call `/hello/world` service. The current logic would search for `/my_namespace/hello/world`

Rayman:
I updated the PR, it now uses the `rospy.resolve_name` function so the logic isn't duplicated. The test suit still runs but maybe we should add some extra tests to it for relative services/topics.

jihoonl:
I think this is right approach to resolve the issue.

@rctoris,  two questions.
1. do you think this approach should be applied to subscriber and publisher too?
2. I think This patch might break many client applications as I assume many users would have used to call global namespaced ros services without leading slash. Where would be nice place to warn or inform users about this change?

jihoonl:
After some testing, it seems fine. Merging.

jihoonl:
Thank @Rayman, @reinzor

PR #184, opened 2015-06-11
Update exception for rosservice_find

dwlee:
`rosservice_find` function raises ROSServiceIOException error when it has communication error with master. 

http://docs.ros.org/jade/api/rosservice/html/rosservice-pysrc.html#rosservice_find


ablakey:
Is raising an ROSServiceIOException not ideal?  What's the case you're running into where you want it to essentially pass on this error?

dwlee:
I think ROSServiceIOException error is handled by application side not rosbridge level. 
So, I closed this PR and will follow up at  https://github.com/RobotWebTools/roslibjs/issues/180

PR #194, opened 2015-09-16
Delete rospy pub/sub instead of unregistering

### LINKED ISSUE #165: Leaked subscriber connections (memory / thread leak)
_This is connected to the rospy issue ros/ros_comm#544 - please see more detailed info there._

There is an intermittent issue where a Subscriber's underlying TCP socket connection thread and objects (`rospy._SubscriberImpl` and `rospy.TCPROSTransport`) can be leaked and never cleaned-up, nor used again when new Subscribers for the same topics are created later on.  This means there is an occasionally occurring, but ever-increasing number of threads running TCP callback loops (end up being NOPs) and leaked memory objects, that increase the memory and usage of the `rosbridge_websocket` server process **_AND**_ the Publisher node on the other side of the TCP connections to which the server is connecting, until the machine runs out of memory and kills one of them.  When this happens with our robot, the kernel oom_killer likes to choose our MotorController process to kill, thus, the issue makes our web-apps using rosbridge somewhat impractical... =/

These leaked connections seem to occur because of a race condition where a new connection (`TCPROSTransport`) is being added to a shared per-topic subscription implementation (`_SubscriberImpl`) that is already closed / being closed. In fact, you can add a debug check in the `_SubscriberImpl.add_connection()` function to see that it is already closed by the time the new connection is attached to it.

Unfortunately, the original, internal use case where this occurs (quite reliably, though still frustratingly intermittently), uses a piece of internal code that interfaces to rosbridge and touches a number of complicated components -- and my attempts to replicate a test case in an parred down, isolated manner just cannot reproduce the issue as reliably as when targeting a running robot and/or using the internal piece of code.  It does seem to happen most when there is a rapid barrage of new Subscribers and unsubscriptions (even, or perhaps specifically to only a limited number of the same topics), and also possibly when the ROS Publisher node that the topics come from is running at high-load.  My attempts to replicate the leak with just standard rospy Subscribers and Publishers (w/o going through rosbridge) have all been unsuccessful so far, but it is hard to produce an exact 1:1 equivalent test program.

The leak itself occurs in rospy, and I have made a ticket there - ros/ros_comm#544.  I believe I even have a "fix" that will catch the symptom of it breaking and prevent the leak-bit; however, _I do not know what is **triggering** the corrupt behavior._  I also do not know if whatever is triggering the improper connection addition/Subscribtion removal at a higher-level is breaking in its own way, or will break after the "fix."  Since I have only seen this in the scenario where rosbridge is triggering the leak, I wanted to post the issue in both places.

I also was not able to get a full understanding on all the pieces where the rosbridge library manages the subscriber objects and callbacks, and some of the code concerned me if the server was using the rospy API in an unexpected way, and thus exposing and triggering the leak.  For instance, could a race condition be being exposed in the explicit deletion of Subscribers, or in the parts that reach down and directly access the `_SubscriberImpl` object instead of using the get_impl() function of the Topic Manager (unlikely here, but for example: `MultiSubscriber.subscribe()` [subscribers.py~L117](https://github.com/RobotWebTools/rosbridge_suite/blob/bf6e9bcc9d6b12386785cb97af6231155deff46f/rosbridge_library/src/rosbridge_library/internal/subscribers.py#L117) ).

Thanks for whatever insight / sanity-checks you have!

mvollrath:
Invoke destructors immediately.

Fixes #165

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/publishers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/publishers.py
@@ -170,7 +170,7 @@ class MultiPublisher():
 
     def unregister(self):
         """ Unregisters the publisher and clears the clients """
-        self.publisher.unregister()
+        del self.publisher
         self.clients.clear()
 
     def verify_type(self, msg_type):
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -93,7 +93,7 @@ class MultiSubscriber():
         self.subscriber = Subscriber(topic, msg_class, self.callback)
 
     def unregister(self):
-        self.subscriber.unregister()
+        del self.subscriber
         with self.lock:
             self.subscriptions.clear()
 


mvollrath:
Gonna keep working on this to get the registrations to disappear.

mvollrath:
Going with this instead: https://github.com/ros/ros_comm/pull/661

PR #197, opened 2015-11-11
Adding Udp 

xuhao1:
I have wrote a UDP protocol based on twisted for rosbridge_server, so that it could work more real-time and I use this protocol for my Unreal-ROS bridge project.

And I have done some simple test.

So if my source is not robost enough, plz let me know.


rctoris:
This looks awesome, thanks for the contribution!!

xuhao1:
Cool

PR #204, opened 2016-01-17
Address Protocol delay_between_messages performance bottleneck

### LINKED ISSUE #203: Protocol delay_between_messages hindering server performance
The Protocol is sleeping for a (default) 10ms after sending each message from the graph to its outgoing method.  This is causing a very severe performance problem.

Affecting code looks like this:  https://github.com/RobotWebTools/rosbridge_suite/blob/d7138dc810aa77b8066a0a2d77195172221a1977/rosbridge_library/src/rosbridge_library/protocol.py#L247-L248

I'm working on a distributed rosbridge app with at least five client connections at the same time, subscribing to some of the same topics, across different hosts.  I tried adding a two-line monkeypatch to `rosbridge_websocket.py`, the gist of it was:

```
from rosbridge_library.protocol import Protocol
Protocol.delay_between_messages = 0
```

This sets the default `delay_between_messages` to zero for new instances of Protocol.  I verified this was being applied with some debug logging in the websocket handler.

The performance boost was tremendous.  I removed all throtte/queue from client code and it ran silky smooth.  I compared this with the default behavior (sleep 10ms after every outgoing) and it showed the signs I'd come to expect from rosbridge under even a small amount of load: messages backing up and slowing down the websocket stream.

The server did use twice as much CPU, but only went from 25-50% aggregate across all threads.  I still expect Tornado to bog down under extreme load, but have yet to hit its limit since the native masking speedups were added to rosbridge_server.

With `delay_between_messages` completely bypassed, the experience of using robridge changes from "I'd better throttle and queue every single topic" to "let 'er rip".

Why is this happening?  I haven't read all of the code, but can prepose that the effect of this delay is multiplied by the number of clients, since a message on one topic (in its own rospy subscription Thread) potentially causes the sleep once per client per message.

There is an undocumented Protocol parameter "message_intervall" that allows for setting `delay_between_messages` per-Protocol instance.  This is not implemented in roslibjs.  https://github.com/RobotWebTools/rosbridge_suite/blob/d7138dc810aa77b8066a0a2d77195172221a1977/rosbridge_library/src/rosbridge_library/protocol.py#L181-L182

I'd like to default `delay_between_messages` to zero, unless there is a compelling case not to.  Selective throttling and queuing works great when you need it.  The only effects of `delay_between_messages > 0` are backlogged, late messages that never get dropped from a queue and a Tornado IOLoop that is sitting around bored while this is happening.

mvollrath:
This prevents performance problems when multiple clients are subscribing to high frequency topics.

Fixes #203

### DIFF
--- a/rosbridge_library/src/rosbridge_library/protocol.py
+++ b/rosbridge_library/src/rosbridge_library/protocol.py
@@ -80,7 +80,7 @@ class Protocol:
     # if this is too low, ("simple")clients network stacks will get flooded (when sending fragments of a huge message..)
     # .. depends on message_size/bandwidth/performance/client_limits/...
     # !! this might be related to (or even be avoided by using) throttle_rate !!
-    delay_between_messages = 0.01
+    delay_between_messages = 0
     # global list of non-ros advertised services
     external_service_list = {}
 


T045T:
LGTM :+1: 

Does anyone else know of a system that would be negatively affected by this change?

pttp:
after reading my comments about this parameter.. i think it should be okay to set this to 0.
If I remember correctly, (slow) clients can request their specific throttling rate anyways, right? 

On January 17, 2016 4:17:14 PM CET, Nils Berg <notifications@github.com> wrote:
>LGTM :+1: 
>
>Does anyone else know of a system that would be negatively affected by
>this change?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/RobotWebTools/rosbridge_suite/pull/204#issuecomment-172335548

-- 
Sent from Mobile Device.

T045T:
For inter-message throttling, yes.
There doesn't seem to be an easy way to set the ```message_intervall``` field in a ```roslibjs``` message, so large, fragmented messages might still overload a slow client, but even then it's still possible to just change the parameter on a per-server basis.

PR #205, opened 2016-01-17
clean up parameters and handling

T045T:
This is a follow-up to #204 

* make parameters accessible via parameter server for all three versions
* remove old advertise_service parameters
* UDP and TCP can't do SSL
* TCP can't authenticate yet (because the RosbridgeTcpSocket class is instantiated for each request and hence does not hold state)
* UDP does not take a hostname or address, but rather an interface

**Note:** I haven't tested these changes yet! I'll be able to do so tomorrow, at least to see whether the launch files still work. I only have access to a system using ```rosbridge_websocket``` not the TCP or UDP version.

### DIFF
--- a/rosbridge_server/launch/rosbridge_tcp.launch
+++ b/rosbridge_server/launch/rosbridge_tcp.launch
@@ -1,61 +1,29 @@
 <launch>
   <arg name="port" default="9090" />
-
   <arg name="host" default="" />
+
   <arg name="incoming_buffer" default="65536" />
   <arg name="socket_timeout" default="10" />
   <arg name="retry_startup_delay" default="5" />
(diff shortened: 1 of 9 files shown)


mvollrath:
This is missing https://github.com/T045T/rosbridge_suite/blob/19af2bf70ef6d4222480d8f8facc3dcc75fb44c0/rosbridge_server/src/rosbridge_server/tcp_handler.py#L35 from https://github.com/T045T/rosbridge_suite/blob/19af2bf70ef6d4222480d8f8facc3dcc75fb44c0/rosbridge_server/src/rosbridge_server/websocket_handler.py#L66

T045T:
udp_handler is, too :O 

I've fixed it, but I'll need to create another pull request, right?

Issue #51, opened 2013-06-10
Defragmentation TODOs


baalexander:
There were a few TODO items in the defragmentation code inside Pull Request #48. I'm moving the items to this ticket for safe keeping and to track the progress.
- [ ] Add integration with rosauth in rosbridge_tcp
- [ ] Adjust max_msg_length value based on further testing in rosbridge_tcp
- [ ] Add documentation
- [ ] Add exception handling
- [ ] Add type checking
- [ ] In defragmentation.py, use String.join() instead of "+=" to improve performance during message reconstruction

Kiloreux:
- [x] In defragmentation.py, use String.join() instead of "+=" to improve performance during message reconstruction
#207

PR #207, opened 2016-01-23
Removed += and add ''.join() to improve performance

Kiloreux:
#51 I removed += in log_msg and exchanged it with ''.join()

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
@@ -90,13 +90,13 @@ class Defragment(Capability, threading.Thread):
                     log_msg = "fragment list " + str(id) + " timed out.."
 
                     if message["id"] != id:
-                        log_msg += " -> removing it.."
+                        log_msg = ''.join([log_msg, " -> removing it.."])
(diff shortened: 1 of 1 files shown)


T045T:
Thanks for the contribution! However, it seems like `+=` might actually be faster than `join()` nowadays:
http://jsperf.com/array-join-vs-string-connect
http://jsperf.com/join-concat/2

pttp:
python: https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation

T045T:
Sorry, I thought this PR was for one of the JavaScript repos, my bad.

After reading that Wiki page, it seems like the patch still contains way more `join()` calls than necessary though.
Shouldn't it only be called when assembling the final string, rather than calling it when adding each part?
Intuitively, the patched version should yield similar performance, since the string is still replaced for each addition.

ablakey:
That would also result in cleaner code. Collect a list of string elements then join them all at the end.

Kiloreux:
@T045T is it okay if i put them in a list and join them by the end ? and squash the 2 commits together ?

T045T:
Sounds good :+1: 
As long as the PR isn't merged, you can squash commits however you like :)

PR #208, opened 2016-01-24
Replaced += with ''.join() for python code

Kiloreux:
I am changing the `log_msg` type from list to string by assignment everytime, shouldn't be a problem beside creating a list and joining it in the end, it should be better this way.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
@@ -87,10 +87,10 @@ class Defragment(Capability, threading.Thread):
                 time_diff = now - self.received_fragments[id]["timestamp_last_append"]
                 if (time_diff.total_seconds() > self.fragment_timeout and
                     not self.received_fragments[id]["is_reconstructing"]):
-                    log_msg = ["fragment list " + str(id) + " timed out.."]
+                    log_msg = ["fragment list ", str(id), " timed out.."]
 
                     if message["id"] != id:
-                        log_msg.extend([" -> removing it.."])
+                        log_msg.append(" -> removing it..")
                         del self.received_fragments[id]
                     else:
                         log_msg.extend([" -> but we're just about to add fragment #"])
@@ -171,14 +171,13 @@ class Defragment(Capability, threading.Thread):
             # Reconstruct the message
             reconstructed_msg = ''.join(self.received_fragments[msg_id]["fragment_list"][0:message["total"]])
             log_msg = ["reconstructed original message:\n"]
-            log_msg.extend(reconstructed_msg)
+            log_msg.append(reconstructed_msg)
             log_msg = ''.join(log_msg)
             self.protocol.log("debug", log_msg)
 
             duration = datetime.now() - now
 
             # Pass the reconstructed message to rosbridge
-            reconstructed_msg = ''.join(reconstructed_msg)
             self.protocol.incoming(reconstructed_msg)
             log_msg = ["reconstructed message (ID:" + str(msg_id) + ") from "]
             log_msg.extend([str(msg_total), " fragments. "])
--- a/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
@@ -87,17 +87,17 @@ class Defragment(Capability, threading.Thread):
                 time_diff = now - self.received_fragments[id]["timestamp_last_append"]
                 if (time_diff.total_seconds() > self.fragment_timeout and
                     not self.received_fragments[id]["is_reconstructing"]):
-                    log_msg = "fragment list " + str(id) + " timed out.."
+                    log_msg = ["fragment list " + str(id) + " timed out.."]
 
                     if message["id"] != id:
-                        log_msg += " -> removing it.."
+                        log_msg.extend([" -> removing it.."])
                         del self.received_fragments[id]
                     else:
-                        log_msg += " -> but we're just about to add fragment #"
-                        log_msg += str(message.get("num")) + " of "
-                        log_msg += str(self.received_fragments[message.get("id")]["total"])
-                        log_msg += " ..keeping the list"
-                    self.protocol.log("warning", log_msg)
+                        log_msg.extend([" -> but we're just about to add fragment #"])
+                        log_msg.extend([str(message.get("num")), " of "])
+                        log_msg.extend([str(self.received_fragments[message.get("id")]["total"])])
+                        log_msg.extend([" ..keeping the list"])
+                    self.protocol.log("warning", ''.join(log_msg))
 
         msg_opcode = message.get("op")
         msg_id = message.get("id")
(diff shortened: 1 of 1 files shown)


rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -87,16 +87,17 @@ def defragment(self, message):
...
                     else:
-                        log_msg += " -> but we're just about to add fragment #"
-                        log_msg += str(message.get("num")) + " of "
-                        log_msg += str(self.received_fragments[message.get("id")]["total"])
-                        log_msg += " ..keeping the list"
+                        log_msg.append(" -> but we're just about to add fragment #")
+                        log_msg.append(str(message.get("num")) + " of ")
+                        log_msg.append(str(self.received_fragments[message.get("id")]["total"]))
+                        log_msg.append(" ..keeping the list")
+                        log_msg = ''.join(log_msg)
```
T045T: This join is indented incorrectly - you're only joining in the `else` clause.
To make it harder to forget, you can join right in the log call:
```
self.protocol.log('warning', ''.join(log_msg))
```

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -135,8 +136,9 @@ def defragment(self, message):
             ):
             self.received_fragments[msg_id]["fragment_list"][msg_num] = msg_data
             self.received_fragments[msg_id]["timestamp_last_append"] = now
-            log_msg = "appended fragment #" + str(msg_num)
-            log_msg += " (total: " + str(msg_total)+ ") to fragment list for messageID " + str(msg_id)
+            log_msg = ["appended fragment #" + str(msg_num)]
+            log_msg.append(" (total: " + str(msg_total)+ ") to fragment list for messageID " + str(msg_id))
+            log_msg = ''.join(log_msg)
```
T045T: For cases like this one, where we're not appending anything conditionally or in a loop, Python string formatting is equally as fast and (in my opinion) a bit easier to read:
```
self.protocol.log(
    'debug',
    'appended fragment #{:d} (total: {:d}) to '
    'fragment list for messageID {:s}'.format(
        msg_num,
        msg_total,
        msg_id)
```

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -168,23 +171,24 @@ def defragment(self, message):
...
-            reconstructed_msg = ""
+            reconstructed_msg = []
             for i in range(0,message["total"]):
-                reconstructed_msg += self.received_fragments[msg_id]["fragment_list"][i]
+                reconstructed_msg.append(self.received_fragments[msg_id]["fragment_list"][i])
 
-            log_msg = "reconstructed original message:\n"
-            log_msg += reconstructed_msg
+            log_msg = ["reconstructed original message:\n"]
+            log_msg.append(reconstructed_msg)
```
T045T: You forgot to join both `log_msg` and `reconstructed_msg` here!

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
Kiloreux: yeah In fact i could do that to all joins and put them directly and save the extra assignement but i guess that would make the code harder to read from my perspective.

T045T:
thanks for the quick changes!
There's one last `join()` on `reconstructed_msg` missing in line 184, the rest looks good to merge :+1:

Kiloreux:
@T045T Thanks for your help, would it be good to merge or there's something else that I need to change ?

T045T:
Apologies, I was on a train the past few hours. Sorry to bother you again, but I do have one more round of comments:
First, I realized that we can skip the for loop for reconstructing the message entirely:
```python
reconstructed_msg = []
for i in range(0,message["total"]):
    reconstructed_msg.append(self.received_fragments[msg_id]["fragment_list"][i])
```
Can become
```python
reconstructed_msg = ''.join(self.received_fragments[msg_id]["fragment_list"][0:message["total"]]
```
since the fragments are already in a list.

Second, for the log messages, technically
```python
log_msg = ["enough/all fragments for messageID ", str(msg_id), " received"]
# use extend() here because join() doesn't like nested lists
log_msg.extend([" [", str(existing_fragments), "]"])
log_msg = ''.join(log_msg)
```
would be faster than
```python
log_msg = ["enough/all fragments for messageID " + str(msg_id) + " received"]
log_msg.append(" [" + str(existing_fragments) + "]")
log_msg = ''.join(log_msg)
```
Since those pluses still do costly concatenation. Not sure if the size/frequency of those log messages is high enough for it to matter, but I thought I'd mention it.

Kiloreux:
The build seems to be failing, it looks like Travis issue, not sure.

T045T:
Yeah, the build failure looks like it's due to broken package dependencies... hopefully that will resolve itself soon.

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -87,17 +87,17 @@ def defragment(self, message):
                 time_diff = now - self.received_fragments[id]["timestamp_last_append"]
                 if (time_diff.total_seconds() > self.fragment_timeout and
                     not self.received_fragments[id]["is_reconstructing"]):
-                    log_msg = "fragment list " + str(id) + " timed out.."
+                    log_msg = ["fragment list " + str(id) + " timed out.."]
```
T045T: you can change these pluses to commas, too :)

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -87,17 +87,17 @@ def defragment(self, message):
...
 
                     if message["id"] != id:
-                        log_msg += " -> removing it.."
+                        log_msg.extend([" -> removing it.."])
```
T045T: When you're just adding a single element, like here, a simple `append()` (without the square brackets) is enough

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -168,23 +169,23 @@ def defragment(self, message):
...
             # Reconstruct the message
-            reconstructed_msg = ""
-            for i in range(0,message["total"]):
-                reconstructed_msg += self.received_fragments[msg_id]["fragment_list"][i]
-
-            log_msg = "reconstructed original message:\n"
-            log_msg += reconstructed_msg
+            reconstructed_msg = ''.join(self.received_fragments[msg_id]["fragment_list"][0:message["total"]])
+            log_msg = ["reconstructed original message:\n"]
+            log_msg.extend(reconstructed_msg)
```
T045T: We can't use extend here, since `reconstructed_msg` is no longer a list, but a string

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -168,23 +169,23 @@ def defragment(self, message):
...
+            log_msg = ''.join(log_msg)
             self.protocol.log("debug", log_msg)
 
             duration = datetime.now() - now
 
             # Pass the reconstructed message to rosbridge
+            reconstructed_msg = ''.join(reconstructed_msg)
```
T045T: Again, `reconstructed_msg` is already a string at this point, no need to `join()` again.

Kiloreux:
@T045T I know I am taking much of your time by checking this every time, sorry.

T045T:
Don't sweat it :)
I was afraid each round of comments might scare you off, thanks for sticking with it!

Kiloreux:
Thanks for all your feedback, really helpful. :smiley:

PR #210, opened 2016-01-29
Abort websocket server listen() retry on shutdown

mvollrath:
This allows the server to shut down via SIGINT or SIGTERM during its listen() retry loop.


jihoonl:
:+1:

PR #219, opened 2016-04-25
Track Twisted run_depend

### LINKED ISSUE #218: Untracked run dependency: Twisted
The UDP handler requires Twisted.  This breaks all the other server scripts (tcp, websocket) if it's not available, because it is imported in the module init:

https://github.com/RobotWebTools/rosbridge_suite/blob/a4a407ae8afe845eec0522efa4cecadccfa009de/rosbridge_server/src/rosbridge_server/__init__.py#L3

Which then raises an `ImportError` when the UDP handler tries to import Twisted:

https://github.com/RobotWebTools/rosbridge_suite/blob/a4a407ae8afe845eec0522efa4cecadccfa009de/rosbridge_server/src/rosbridge_server/udp_handler.py#L6

`ImportError: No module named twisted.internet.protocol`

Run dependencies need to be tracked in package.xml.  In Indigo on Ubuntu trusty, the `python-twisted-core` package is sufficient to suppress this error, but I can't say whether or not the UDP handler will work against this version of Twisted.

Fortunately `python-twisted-core` is in the rosdep tree and this is a one-line fix:

https://github.com/ros/rosdistro/blob/17fa310173bc4cd4a9b240f44f9fe3f976dba777/rosdep/python.yaml#L2209-L2221

mvollrath:
Fixes #218

### DIFF
--- a/rosbridge_server/package.xml
+++ b/rosbridge_server/package.xml
@@ -18,6 +18,7 @@
   <build_depend>rosbridge_library</build_depend>
   <build_depend>rosapi</build_depend>
   <build_depend>rospy</build_depend>
+  <run_depend>python-twisted-core</run_depend>
   <run_depend>rosbridge_library</run_depend>
   <run_depend>rosapi</run_depend>
   <run_depend>rospy</run_depend>


mvollrath:
Thanks @rctoris any chance of updated .debs for this?

rctoris:
PR's submitted!

https://github.com/ros/rosdistro/pull/11291
https://github.com/ros/rosdistro/pull/11292

k-okada:
unfortunately  0.7.14 does not incldues this fix, please release updated version. ;-)
 https://github.com/RobotWebTools/rosbridge_suite/commits/0.7.14

PR #221, opened 2016-04-26
Fix UDP handler shutdown

mvollrath:
* Update `.gitignore`
* Isolate rosbridge_server libraries from unnecessary imports
* Stop UDP server on ROS shutdown

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.pyc
 *.iml
+*.swp
 .project
 .pydevproject
 .DS_Store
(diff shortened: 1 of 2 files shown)


jihoonl:
Looks good. and one question. Can you give us some more information [the second commit](https://github.com/RobotWebTools/rosbridge_suite/pull/221/commits/50dd4f3affe60883f1981d316d7f6fc11f73ab95)? Does it behave differently?

jihoonl:
@mvollrath Could you explain a bit more about the second change? My worries is that this change might break other packages which depend on `rosbridge_server` as library.

mvollrath:
The second change would break other packages which depend on `rosbridge_server` as a library.  I'll take it out and leave the UDP handler fix if that's no good.

jihoonl:
@mvollrath sounds good thank you very much :). As soon as you make the change, I will merge it.

mvollrath:
@jihoonl this was done.

jihoonl:
Sorry buddy. I am merging it now. Thanks!

PR #229, opened 2016-07-03
Fixed deprecated code in pillow

vladrotea:
fromstring() method was completely removed since pillow 3.0.0 (https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#300-2015-10-01) and now has a direct replacement.


jihoonl:
:+1:

PR #234, opened 2016-08-09
New service and types of topics

marcoarruda:
UPDATED

- Get node details: subscriptions, publishing topics and services

There's also a pull request RosLibJS side for those services
https://github.com/RobotWebTools/roslibjs/pull/234


jihoonl:
:+1:

PR #238, opened 2016-10-13
Added new parameters for topic and service security.

MasterEric:
This pull request adds three new parameters to both rosapi and rosbridge_server. These three parameters, arrays titled _topics_glob_, _services_glob_, and _params_glob_, allow you to control which topics, services, and parameters, respectively, are broadcast by rosbridge_server, thus preventing the end user from modifying topics they should not be allowed to.


T045T:
In addition to the review:
As far as I see, there's logging for failed (and successful) glob matches on the server side.
I'm not sure how far along support for [Status Messages](https://github.com/RobotWebTools/rosbridge_suite/blob/develop/ROSBRIDGE_PROTOCOL.md#32-status-messages) is, but if they're working, we could also provide feedback to the browser on failed matches, in case the mismatch is a mistake, rather than malicious.

kmhallen:
@T045T Great feedback. We will work to implement your suggestions, but it might be up to two weeks.

MasterEric:
Pushed my fixes and updates.

I made a new pull request for the time fix, but I'm not sure how to remove that change from the commits on this branch.

rosapi/src/rosapi/params.py
```diff
@@ -42,43 +43,82 @@
...
-    
-def get_param(name, default):
-    d = None
-    if default is not "":
+def set_param(name, value, params_glob):
+    if params_glob and not any(fnmatch.fnmatch(str(name), glob) for glob in params_glob):
+        # If the glob list is not empty and there are no glob matches,
+        # stop the attempt to set the parameter.
+        return
+    else:
```
T045T: You can drop the else (and subsequent indentation here, because you return in the if's body above. That should help Github see that you didn't actually change all the lines in the else part :)

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -125,6 +139,48 @@ def shutdown_hook():
             print "--max_message_size argument provided without a value. (can be None or <Integer>)"
             sys.exit(-1)
 
+    if "--topics_glob" in sys.argv:
+        idx = sys.argv.index("--topics_glob") + 1
+        if idx < len(sys.argv):
+            value = sys.argv[idx]
+            if value == "None":
+                RosbridgeWebSocket.topics_glob = None
```
T045T: Looks like RosbridgeWebSocket.topics_glob and RosbridgeWebSocket.services_glob aren't used anywhere else, except below, to initialize the Capabilities - is there a reason you set it?

rosbridge_library/src/rosbridge_library/capabilities/publish.py
```diff
@@ -56,6 +59,20 @@ def publish(self, message):
         latch = message.get("latch", False)
         queue_size = message.get("queue_size", 100)
 
+        if self.topics_glob:
```
T045T: I think technically you want ```Publish.topics_glob``` here (and ```ClassName.foo_glob``` in other places).
```self.topics_glob``` implicitly copies the value of ```Publish.topics_glob``` the first time you access it, but you can set it independently afterwards.

MasterEric:
More fixes and updates pushed.

T045T:
Thanks for your work, looks great now!

Issue #240, opened 2016-10-26
Type mismatch (int64 & long)


AlexisTM:
Dear,

I have a type mismatch between `roslibjs` and `rosbridge`. I don't know if the issue is related to `rosbridge` or `roslibjs`. 

I am calling a service containing a `Task.msg` defined by : 

**TaskHandle.srv**
```
Task task
---
string message
bool result
```

**Task.msg**
```
string name                 # Can be unique, you can use it as you please
uint8 mission_type          # Type of the mission
geometry_msgs/Point position
float64 yaw                 # in degrees
float32[] data              # Can be multiple data, like time to sleep, time to loiter, etc.
int64 ID
```

The error message I get is : 

```
[ERROR] [WallTime: 1477462156.860756] [Client 2] [id: call_service:/flyingros/controller/task/remove:6] call_service FieldTypeMismatchException: flyingros_msgs/TaskHandleRequest message requires a int64 for field task.ID, but got a <type 'long'>
```

task.ID is an `int64` which **should be compatible** with the type long in python according to http://wiki.ros.org/msg#line-47

The problem comes from definitions in the message conversion  where "int64" should be declared as long. 

https://github.com/RobotWebTools/rosbridge_suite/blob/ee339d54474b87aa2f56b4755cbf095a99d25bff/rosbridge_library/src/rosbridge_library/internal/message_conversion.py#L53

T045T:
Merged your PR, thanks again!

PR #241, opened 2016-10-26
Type conversion convention correction, correcting issue #240

AlexisTM:
Correction issue #240, according to the [ROS message convention](http://wiki.ros.org/msg#line-47) for the int64 type.


T045T:
Thanks for fixing this!

PR #245, opened 2016-12-09
New rosapi service: get actionlib servers

marcoarruda:
New rosapi service: get actionlib servers.
For a new widget that I'm developing


jihoonl:
+1

PR #247, opened 2017-01-12
Delay unregister to mitigate !138

v-lopez:


T045T:
Looks like some tests are failing because they expect deregistration to be instant - can you take a look at those and fix them, please?

v-lopez:
@T045T I have modified the tests. Due to the nature of the changes I have added to add sleeps matching the duration of the unregister timeout (currently 10seconds) so the tests need 2-3 minutes to run instead of 1. If that is a problem, the duration of the timeout can be reduced on the unit tests, but I'd rather test the code without any modifications.

T045T:
Thanks!
I think it's fine to set unregister_timeout to something smaller for testing, rather than triple the test execution time.
One more thing: Since it's a member now, can you change ```UNREGISTER_TIMEOUT``` to not be all caps?

v-lopez:
Sure if we're not going to treat it as a constant either it's fine.

T045T:
Class-level constants are somewhat unusual, so I'd prefer to either make it a class member (-> lowercase name) or a module constant ( -> uppercase name). In the latter case, it's still possible to access (and modify) it from test code (looks like that was the reason you changed it into a member in the first place).

Since we don't really expect users to mess with it, a module constant is probably the way to go.

T045T:
Great work, thank you!

PR #248, opened 2017-01-13
Fix: Set default to publish all topics

plieningerweb:
Without better doc, one does not understand why no topics are published. I thought, something is broken.
With this defaults, everything is working out of the box. And for a more secure setup, one can change it.


jihoonl:
:+1:

T045T:
I agree that missing doc is to blame, but the bug wasn't the default values, but rather the fact that an empty list did not get interpreted as "don't use globs" (see 1112937). I'll look into this and add some documentation, too.

Issue #253, opened 2017-01-18
BSON mode for bidirectional communication


Sanic:
After reading https://github.com/RobotWebTools/rosbridge_suite/issues/169 and a couple of related Issues, i saw that there is already some progress on using BSON for large amounts of (binary) data. However, if i understand this correctly, you can only use BSON for binary data flowing from ROS **to** the rosbridge Client. 
In my usecase, i need to transfer large amounts of binary data from the rosbridge client to the ROS infrastructure. Additionally, i would like to send data from the rosbridge client side without encoding it in base64, which leads to an overhead in terms of computational power and transmission bandwidth.

Is it possible to add a _bson-only-mode_ to the server that replaces all JSON encoding/decoding with BSON without rewriting large amounts of the core?
By doing that, we could still use JSON for legacy or "non-binary" applications, where as BSON can be used for scenarios where large amounts of binary data needs to be transferred.
When switch the server to a _bson-only-mode_, we also don't have to incorporate any changes to the ROSBridge protocol itself as far as i can tell (with my limited knowledge of the server internals).

/edit: Using BSON might also improve the stable receipt of messages, because it's prefixed with the total message length in bytes and terminates by a fixed 0x0 at the end. This might lead to an improvement for discussions like in https://github.com/RobotWebTools/rosbridge_suite/issues/140 or https://github.com/RobotWebTools/roslibjs/issues/126

Sanic:
This issue can be closed now, due to #257 
There is also a PR for BSON in Websockets #327 , which would cover the most use cases.

PR #257, opened 2017-01-28
Implemented a bson_only_mode flag for the TCP version of rosbridge

Sanic:
Related to https://github.com/RobotWebTools/rosbridge_suite/issues/253

This commit allows users to switch to a full-duplex transmission of BSON messages and therefore eliminates the need for a base64 encoding of binary data when running in this mode. 
The new mode is **fully optional** and **is not activated by default**.
You can use the new mode by starting:
- roslaunch rosbridge_server rosbridge_tcp.launch bson_only_mode:=True

or
- rosbridge_tcp.py --bson_only_mode 

Your rosbridge_server will now only expect and transmit data in BSON.

This PR can also provide an improvement for problems like in https://github.com/RobotWebTools/roslibjs/issues/126, 
because the TCP handler can make some assumptions on the length of the message when using BSON. This is due to the length field that is located in the first four bytes of every BSON message.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -70,11 +70,13 @@ binary_encoder = None
 def get_encoder():
     global binary_encoder
     if binary_encoder is None:
-        binary_encoder_type = rospy.get_param('~binary_encoder', 'b64')
-        if binary_encoder_type == 'b64':
-            binary_encoder = standard_b64encode
-        elif binary_encoder_type == 'bson':
-            binary_encoder = bson.Binary   
+        binary_encoder_type = rospy.get_param('~binary_encoder', 'default')
+        bson_only_mode = rospy.get_param('~bson_only_mode', False)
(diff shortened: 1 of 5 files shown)


jihoonl:
hi @Sanic, it looks actually very nice. Do you have any test client example that we can test? Or it would be very nice if you can provide us test programs too.

Since I don't have any implementation of tcp client, I am not sure how to test this updates.

Sanic:
Hi @jihoonl 
To keep things simple i've created a small test client for you.
Since rosbridge is written in python, i also implemented the client in python.
You can find the client over here: https://github.com/Sanic/ROSBridgeTestclient

The client advertises the topic /bson_test , publishes a message to it and calls a service afterwards. 
The service response will be printed in the test client.

To get everything running, i put a description in the README.md of the repository. I would recommend to use TEST_MODE=JSON (default) first, to see if everything is fine with the normal JSON transmission in the first place.

Please let me know if there are any issues.

jihoonl:
Nice thanks!
:+1:

Issue #198, opened 2015-11-25
Rosbridge only one way on OSX


cfry:
This issue is more fully described by:
http://answers.ros.org/question/221356/rosbridge-only-one-way-on-my-osx/

In a nutshell, I can run the tutorial http://wiki.ros.org/roslibjs/Tutorials/BasicRosFunctionality
with messages successfully leaving my browser, reaching the tutorial server thru rosbridge,
and successfully computing the answer (according to printouts in the terminal window) , but nothing comes back to the browser.  (the callback isn't called and doesn't look like any web  messages
are coming in either according to Chrome Dev Tools.)
No errors occur that I can see. 
I can run the turtlesim demo fine and overall ROS seems to be behaving well other than this problem.

From the above ROS answers URL, you can see that commenter Philipp was able to run this
tutorial successfully on his Ubuntu machine and that he had similar 
ROS environment variables and /etc/hosts as me. 
Answerer William Woodall was able to reproduce this problem on his OSX Yosemite & Indigo machine
(same as mine) and suggested I open this issue in rosbridge. 
Comments in his answer sound like good clues to start with.

NOTE I would be happy if this issue was already fixed in Jade, but I have not tested it in Jade as
installing ROS on OSX is, um, difficult for me, but worth the switch if that's what it takes to resolve
this as its a showstopper for my app.

PS I'm running Python 2.7 and downloaded ROS indigo and Rosbridge about 2 weeks ago.
THANK YOU!

T045T:
I'll try to get Indigo set up on my MacBook and reproduce this for testing by this weekend - sorry, I won't get to it any earlier. Any others with OS X machines?

A couple of questions:
* Are you using the system Python or a brewed version?
* In your Answers post, you mention that rosbridge was difficult to install - can you write down the steps you took to install it? Maybe they'll get us started on brew support for rosbridge :)

jonfink:
Also having this same issue! 
- Using system python
- rosbridge install was smooth for me (cloned rosauth and rosbridge_suite git repos into a workspace)

I've started to dig in to see where messages (published from ROS) stop moving forward but I've only gotten so far as to determine that `Protocol.send` is being called but the output of `self.serialize` is `None`

jonfink:
bson dependency turned out to be the issue. Doing a `pip insatll pymongo` (which includes it's own bson module) and things seem to work for me now.

cfry:
Three Cheers! 
pip install pymongo
solved  the problem.
[...]

PR #270, opened 2017-03-10
Check if appropriate bson module is installed or not just after init_node (#198)

ledmonster:
Check whether mongodb oriented bson is installed or not just after init_node,
and if appropriate bson is not installed, shutdown node with error message.

See: https://github.com/RobotWebTools/rosbridge_suite/issues/198


cjds:
This wrong BSON library thing tripped me up for quite a while.

jihoonl:
ping

jihoonl:
@ledmonster sorry I just realized that all reviews were in pending. Could you change the requested in review?

jihoonl:
Looks good to me. thanks. merging it now.

PR #273, opened 2017-03-14
Set default for bson_only_mode in websocket handler and launch file.

Sanic:
This PR shall prevent bson_only_mode to be undefined at any time when using the websocket version of ROSBridge. This PR targets the problem described in https://github.com/RobotWebTools/rosbridge_suite/issues/271 .

When this change seems fine, i could also apply it on the udp version of ROSBridge.
Is there any UDP testclient?


jihoonl:
Thanks. I don't have any udp example on my hand. If anyone has it, leave a comment please :)

PR #276, opened 2017-04-06
Document glob

sevenbitbyte:
There was a breaking change introduced by PR #238 

This change now requires all rosapi users to configure their globs to allow any topic/service/param information to be returned.

I've added a README which documents these new parameters along with two example launch configurations.

Please let me know if I've understood this feature correctly.

@MasterEric @T045T

### DIFF
--- a/rosapi/README.md
+++ b/rosapi/README.md
@@ -1,0 +1,11 @@
+rosapi
+===============
+
+#### Parameters
+
+ `~topics_glob` (string, default '')
+ `~services_glob` (string, default '')
+ `~params_glob` (string, default '')
+ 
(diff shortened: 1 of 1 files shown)


T045T:
Your changes do describe the way these parameters currently work. Thanks for adding the documentation!

The launch files we ship have been changed to include the default values your example mentions, which might be worth mentioning.

I've also been thinking that it would be nice to have an option to deactivate the globbing mechanism altogether if performance is more of a concern than security, but that's probably a discussion for another time.

T045T:
I also think it's good idea to remind people to whitelist the ```rosapi``` namespace.

jihoonl:
The PR itself still looks good. I would merge this and make follow-up jobs to improve what @T045T mentioned. I have created #279 to document roslaunch configuration and rosapi whitelisting.

PR #277, opened 2017-04-06
don't try to set TCP nodelay option for UDP

### LINKED ISSUE #275: issue with a set_nodelay variable in  udp_handler.py
rosbridge_udp.launch crashes with the error at line 47 in  udp_handler.py:   `self.set_nodelay(True)` 
Changing that to self.set_nodelay=True fixes the issue!

T045T:
fixes #275


jihoonl:
:+1:

PR #287, opened 2017-08-24
Fixing errors in test.

jihoonl:
* add timeout and a proper service callback wait-logic instead of sleep.
* add queue_size in rospy publisher


jihoonl:
I have double checked this patch fixes travis error.

PR #252, opened 2017-01-17
Bug fix to rosapi: allow empty lists and set default all for all launch files

TheDash:

PR #292, opened 2017-08-30
Removes array delimiters while parsing parameters

### LINKED ISSUE #291: Service name matching is not working
I have a service in the root name space called `get_errors` and the ros_bridge_library fails to recognize the service unless I pre-pend a `/` to to the name in `services_glob` parameter
I'm getting an error `[Client 0] No match found for service, cancelling service call...`

Behery:
Also refactors the way the parameters are obtained. Fixes #291

### DIFF
--- a/rosapi/scripts/rosapi_node
+++ b/rosapi/scripts/rosapi_node
@@ -45,18 +45,18 @@ def get_globs():
     global topics_glob
     global services_glob
     global params_glob
-    topics_glob = [
-        element.strip().strip("'")
-        for element in rospy.get_param('~topics_glob', '')[1:-1].split(',')
-        if len(element.strip().strip("'")) > 0]
-    services_glob = [
-        element.strip().strip("'")
-        for element in rospy.get_param('~services_glob', '')[1:-1].split(',')
-        if len(element.strip().strip("'")) > 0]
-    params_glob = [
-        element.strip().strip("'")
-        for element in rospy.get_param('~params_glob', '')[1:-1].split(',')
-        if len(element.strip().strip("'")) > 0]
+
+    def get_param(param_name):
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
The Python changes look good.  I like the refactor: one local function is much easier to reason about and maintain than three list comprehensions. It's also good that this only strips array brackets rather than truncating the first and last character and implicitly assuming array brackets are present.

There's a part of me that wants to find a way to perform these string manipulations all in one pass using a regular expression capture but  that's an indulgence worth ignoring unless this actually becomes a performance problem.

There are no tests for this project but can you provide a minimal example that is broken with the latest release and fixed by this PR?

Thanks!

nuclearsandwich:
Oh, forgot to mention, the changes that this complements / fixes were introduced in https://github.com/RobotWebTools/rosbridge_suite/pull/252

Behery:
Using the following launch file would cause it to break in the current release. The rosbridge_server will try to match service requests against 'o_something', and 'do_something_els' and will not be able to call the correct services.

```
<launch>
    <node name="my_ros_bridge_node" pkg="rosbridge_server" type="rosbridge_websocket.py" cwd="node" output="screen" respawn="true">
        <param name='services_glob' value='do_things, do_something_else'/>
    </node>
</launch>
```

nuclearsandwich:
@jihoonl do we have a merge strategy for pull requests beyond "merge when ready"? Is there anything I should do to merge this, like changelog updates or other notes?

jihoonl:
Not really. As soon as you approve, feel free to merge. I am merging it.

nuclearsandwich:
Great! Thanks Behery for the contribution 🌈

PR #297, opened 2017-09-28
Handles empty globes properly

### LINKED ISSUE #296: if globs aren't set then the rosapi's services aren't fully functional
If the globs for services and topics aren't set, then the following services in rosapi return an empty list:
```
/rosapi/services
/rosapi/topics
/rosapi/action_servers
/rosapi/get_services_for_type
/rosapi/service_node
...
```
and probably more

Behery:
The api is now able to handle empty globs. Meaning that the service calls are now functional even if all the topics/services/params are exposed. Fixes #296

### DIFF
--- a/rosapi/src/rosapi/proxy.py
+++ b/rosapi/src/rosapi/proxy.py
@@ -90,44 +90,48 @@ def get_nodes():
     """ Returns a list of all the nodes registered in the ROS system """
     return rosnode.get_node_names()
 
+
 def get_node_publications(node):
     """ Returns a list of topic names that are been published by the specified node """
     try:
-      publishers, subscribers, services = Master('/rosbridge').getSystemState()
-      toReturn = []
-      for i,v in publishers:
-        if node in v:
-          toReturn.append(i)
-      toReturn.sort()
-      return toReturn
+        publishers, subscribers, services = Master('/rosbridge').getSystemState()
+        toReturn = []
+        for i, v in publishers:
+            if node in v:
+                toReturn.append(i)
+        toReturn.sort()
+        return toReturn
     except socket.error:
-      return []
-    
+        return []
+
+
 def get_node_subscriptions(node):
     """ Returns a list of topic names that are been subscribed by the specified node """
     try:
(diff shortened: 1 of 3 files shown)


rosapi/src/rosapi/proxy.py
```diff
@@ -150,21 +159,31 @@ def filter_action_servers(topics):
...
+            if(possible_action_server == namespace and topic == "cancel"):
+                possibility[0] = 1
+            if(possible_action_server == namespace and topic == "feedback"):
+                possibility[1] = 1
+            if(possible_action_server == namespace and topic == "goal"):
+                possibility[2] = 1
+            if(possible_action_server == namespace and topic == "result"):
+                possibility[3] = 1
+            if(possible_action_server == namespace and topic == "status"):
+                possibility[4] = 1
```
jihoonl: Thanks for fixing this method too.
`filter_action_servers` method can be simplified like this way. Could you make this change?

```
    possibility = [0, 0, 0, 0, 0]
    action_topics = ['cancel', 'feedback', 'goal', 'result', 'status']
    for topic in sorted(topics):
        if (len(topic.split('/')) == 3):
            [empty, namespace, topic] = topic.split('/')
            if(possible_action_server != namespace):
                possible_action_server = namespace
                possibility = [0, 0, 0, 0, 0]
            if possible_action_server == namespace and topic in action_topics:
                possibility[action_topics.index(topic)] = 1
        if all(p == 1 for p in possibility):
            action_servers.append(possible_action_server)
```

rosapi/src/rosapi/proxy.py
```diff
@@ -45,16 +45,20 @@
 
 from rosapi.msg import TypeDef
 
+from glob_helper import filter_globs, any_match
```
jihoonl: Use `from .glob_helper import filter_globs, any_match` to import from local module.

Behery:
Hi @jihoonl could you please check the latest updates?
Thanks

jihoonl:
Thanks.

PR #300, opened 2017-11-06
Add Python3 compatibility

kartikmohta:
All the unit tests seem to pass with both Python 2 and 3.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
@@ -34,12 +34,8 @@ import fnmatch
 from functools import partial
 from rosbridge_library.capability import Capability
 from rosbridge_library.internal.services import ServiceCaller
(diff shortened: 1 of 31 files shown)


rosbridge_library/src/rosbridge_library/capabilities/call_service.py
```diff
@@ -35,12 +35,17 @@
 from rosbridge_library.capability import Capability
 from rosbridge_library.internal.services import ServiceCaller
 
+import sys
+if sys.version_info >= (3, 0):
+    string_types = str
+else:
+    string_types = (str, unicode)
```
jihoonl: Can you move this into `rosbridge_library/utils/__init__.py` and import from there? Then, we can avoid copy and pasting of the same functionality.

kartikmohta: Yes, that seems better 👍

PR #305, opened 2017-12-07
Fixed action_servers filter to allow more than one namespace

milesial:
### Context
The service `/rosapi/action_servers` uses a filter function on topic names to detect action servers.

### Problem
This function worked only with topics of type `/namespace/goal`, with only one namespace.

### Fix
It is now capable of detecting for example `/ns1/ns2/server/goal` (when all goal, success, cancel, feedback and status topics are alive, of course).
The service returns an array of the full namespaces, `/ns1/ns2/server` being one of them here for example


jihoonl:
Thanks for the help. @Behery

jihoonl:
And thanks for the patch @milesial :)

PR #312, opened 2018-01-08
Service fixes

### LINKED ISSUE #265: High CPU load when services advertised by clients are called but dont deliver service_response
Hi everyone,

maybe this is working as intended, but I noticed that rosbridge has permanently high CPU load after forwarding call_service messages that are not answered.

My test client sends the following messages and then disconnects from rosbridge_websocket:
~~~
# Advertise service
{"type": "std_srvs/SetBool", "service": "/horse/test_service", "op": "advertise_service"}

# call the service
{"args": {"data": true}, "service": "/horse/test_service", "op": "call_service"}

# rosbridge forwards the message like this:
{"args": {"data": true}, "id": "service_request:/horse/test_service:1", "service": "/horse/test_service", "op": "call_service"}

# client unadvertises the service
{"service": "/horse/test_service", "op": "unadvertise_service"}
~~~

In the end rosbridge_websocket continues to block one CPU core, presumably with busy waiting:
~~~
ps aux | grep rosbridge
juelg    30485 81.6  0.1 741984 58468 ?        Ssl  19:20   0:59 python /home/juelg/robot_folders/checkout/horse_rosbridge_suite/catkin_workspace/src/rosbridge_suite/rosbridge_server/scripts/rosbridge_websocket __name:=rosbridge_websocket __log:=/home/juelg/.ros/log/fed3a90e-f85d-11e6-9556-c85b766c783b/rosbridge_websocket-1.log
~~~

Is this the expected behavior or a bug?
I tested against the bleeding edge "develop" branch as well as against 0.7.16.

T045T:
Primarily, this should fix #265 (but only if a service is actually unadvertised. If it just loses connection and does not answer, the client will still busy-wait).

I added new rostests to confirm the fix works, and discovered a few minor bugs in the process. Since these changes are only related via the rostest, I'd prefer not to squash-merge this PR.

### DIFF
--- a/rosbridge_library/test/capabilities/test_service_capabilities.py
+++ b/rosbridge_library/test/capabilities/test_service_capabilities.py
@@ -1,0 +1,136 @@
+#!/usr/bin/env python
+import rospy
+import rostest
+import unittest
+
+from json import loads, dumps
+
+from rosbridge_library.capabilities.advertise_service import AdvertiseService
+from rosbridge_library.capabilities.unadvertise_service import UnadvertiseService
+from rosbridge_library.capabilities.call_service import CallService
+from rosbridge_library.capabilities.service_response import ServiceResponse
+from rosbridge_library.protocol import Protocol
+from rosbridge_library.protocol import InvalidArgumentException, MissingArgumentException
+
+
+class TestServiceCapabilities(unittest.TestCase):
+    def setUp(self):
+        self.proto = Protocol(self._testMethodName)
+        # change the log function so we can verify errors are logged
+        self.proto.log = self.mock_log
+        # change the send callback so we can access the rosbridge messages
+        # being sent
+        self.proto.send = self.local_send_cb
+        self.advertise = AdvertiseService(self.proto)
+        self.unadvertise = UnadvertiseService(self.proto)
+        self.response = ServiceResponse(self.proto)
+        self.received_message = None
+        self.log_entries = []
+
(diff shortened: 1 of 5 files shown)


jihoonl:
Can you clean up commits if you prefer rebase and merge? For curiosity, why you don't want to squash-merge for rostest related changes?

jihoonl:
@Behery  can you review the changes?

Behery:
sure .. I'll take a look in the afternoon today

T045T:
I think these commits are about as clean as they can get - I can squash the typo fix into some other commit, but I'd prefer the history to clearly show when and why each change was made. I could also break up the rostest commit and add the ```graceful_shutdown``` tests in the commit I added the feature, and the other ones with the other fixes. Let me know what you think :)

T045T:
I've shuffled around the code a bit, hopefully this is clearer :)

I checked, and both commits leave the repository in a state where it builds and passes all tests.

rosbridge_library/test/capabilities/test_service_capabilities.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python
+from threading import Thread
+
+import rospy
+import rostest
+import unittest
+
+from json import loads, dumps
+from std_msgs.msg import String
```
Behery: Unused import, sorry i must've missed this the last time

T045T:
Done, good catch!

PR #314, opened 2018-01-09
Fixes #313 by fixing has_binary in protocol.py

### LINKED ISSUE #313: Serializing a list containing binaries doesn't work
When the `rosbridge_library.protocol` tries to serialize an object, the has_binary function doesn't check for list of binaries; it only checks if the object itself is binary, or if it's a `dict` that has binary in it.
This caused the serialize function to try and json serialize it (which would fail in that case) and the service call doesn't respond back to the client.

Behery:
Checks for lists of binary objects inside the objects as well as dictionaries


Behery:
I think something went wrong with my fork .. i'll create a cleaner pull request

PR #315, opened 2018-01-09
Fixes #313 by fixing has_binary in protocol.py

### LINKED ISSUE #313: Serializing a list containing binaries doesn't work

Behery:
Checks for lists that have binary content as well as dicts

### DIFF
--- a/rosbridge_library/src/rosbridge_library/protocol.py
+++ b/rosbridge_library/src/rosbridge_library/protocol.py
@@ -32,10 +32,10 @@
 
 import rospy
 import time
+
 from rosbridge_library.internal.exceptions import InvalidArgumentException
 from rosbridge_library.internal.exceptions import MissingArgumentException
 
-#from rosbridge_library.internal.pngcompression import encode
 from rosbridge_library.capabilities.fragmentation import Fragmentation
 from rosbridge_library.util import json, bson
 
@@ -52,19 +52,13 @@ def has_binary(obj):
     """ Returns True if obj is a binary or contains a binary attribute
     """
 
-    if type(obj) is bson.binary.Binary:
-        return True
+    if isinstance(obj, list):
+        return any(has_binary(item) for item in obj)
 
-    if type(obj) is list:
-        for item in obj:
-            if has_binary(item):
-                return True
+    if isinstance(obj, dict):
+        return any(as_binary(item) for item in obj.itervalues())
 
-    if type(obj) is dict:
-        for k, v in obj.iteritems():
-            if has_binary(v):
-                return True
-    return False
+    return isinstance(obj, bson.binary.Binary)
 
 
 class Protocol:
--- a/rosbridge_library/src/rosbridge_library/protocol.py
+++ b/rosbridge_library/src/rosbridge_library/protocol.py
@@ -46,15 +46,26 @@ def is_number(s):
         return True
     except ValueError:
         return False
-        
-def has_binary(d):
-    if type(d)==bson.Binary:
+
+
+def has_binary(obj):
+    """ Returns True if obj is a binary or contains a binary attribute
+    """
+
+    if type(obj) is bson.binary.Binary:
         return True
-    if type(d)==dict:
-        for k,v in d.items():
+
+    if type(obj) is list:
+        for item in obj:
+            if has_binary(item):
+                return True
+
+    if type(obj) is dict:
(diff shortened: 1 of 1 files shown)


Behery:
@jihoonl can you please review this?

jihoonl:
Ok. Please wait for me until I get the ubuntu machine.. it is coming

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -46,15 +46,26 @@ def is_number(s):
...
-    if type(d)==dict:
-        for k,v in d.items():
+
+    if type(obj) is list:
+        for item in obj:
+            if has_binary(item):
+                return True
+
+    if type(obj) is dict:
+        for k, v in obj.iteritems():
```
T045T: you don't use `k` here, so you can just do `for item in obj.itervalues()`

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -46,15 +46,26 @@ def is_number(s):
...
             if has_binary(v):
```
T045T: You could simplify both this and the `list` case using `any()` and a generator expression:
```
# list case
return any((has_binary(item) for item in obj))
# dict case
return any((has_binary(item) for item in obj.itervalues()))
```

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -46,15 +46,26 @@ def is_number(s):
...
-        
-def has_binary(d):
-    if type(d)==bson.Binary:
+
+
+def has_binary(obj):
+    """ Returns True if obj is a binary or contains a binary attribute
+    """
+
+    if type(obj) is bson.binary.Binary:
```
T045T: Throughout: Please use `isinstance()` instead of `type()` (see https://stereochro.me/ideas/type-vs-isinstance)

jihoonl: +1

rosbridge_library/src/rosbridge_library/protocol.py
jihoonl: +1

Behery:
Thanks guys :)

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -32,10 +32,7 @@
 
 import rospy
 import time
-from rosbridge_library.internal.exceptions import InvalidArgumentException
```
T045T: These imports are important - they hoist the exceptions from the `internal.exceptions` namespace to `rosbridge_library.protocol` (see https://travis-ci.org/RobotWebTools/rosbridge_suite/jobs/329044753)

Behery: mmm ... i thought they were placed in the tests by mistake or so ... i'll return them in the morning then 👍

Behery: Just out of curiosity, why do we need to do that? Why not just import them from internal.exceptions?

T045T: Well, I'm not sure we absolutely need to, but removing those imports now would be a breaking API change in any case, which means a major version bump.

Behery: i think they're imported from protocol only in the tests ... i did a quick search in the repo, and it seems only the tests use them like that.

T045T: In our code, yes. I'm talking about external code that uses `rosbridge_library` to implement some sort of custom bridge. There's not going to be a lot of that, but as I said, we'd be changing our public API. Let's keep the imports for now, and make a note somewhere to keep them in mind for when there's more reasons to justify a major version bump :)

Behery: yes ... i returned them for now. If we need to remove them later then it should be in another PR anyways. 
Thanks for the explanation :)

Issue #162, opened 2015-01-05
find out how to use tornado as submodule of rosbridge_tools


jihoonl:
For better uses,

Related PR & issues 
- https://github.com/RobotWebTools/rosbridge_suite/pull/157
- https://github.com/RobotWebTools/rosbridge_suite/issues/154
- https://github.com/RobotWebTools/rosbridge_suite/pull/155

Branchs
- https://github.com/RobotWebTools/rosbridge_suite/tree/rosbridge_tools_update
- https://github.com/RobotWebTools/rosbridge_suite/tree/tornado_magic_076

jihoonl:
As reference,

rocon uses a fork or redis-py as submodule of rocon_python_redis package. The local changes are necessary to use as submodule.

**Example**
Original : https://github.com/andymccurdy/redis-py/blob/2.6.2/redis/__init__.py
Rocon fork : https://github.com/robotics-in-concert/rocon_tools/blob/indigo/rocon_python_redis/src/rocon_python_redis/__init__.py

jihoonl:
Should we keep the tornado in rosbridge_server in kinetic or separate it out?

asmodehn:
Please separate it out :)

I recently stumbled on this, while checking dependencies for https://github.com/asmodehn/celeros (and eventually https://github.com/asmodehn/rostful)
I had a tornado python package installed in ros indigo, and I had no idea where it came from for a long while.

The system (ubuntu) tornado package is also too old for me (I need >3.2), so I would need a tornado dependency as ros package. Following the ros distribution strategy, we should find a version that work for all projects that use it, and turn it into a ros package...

We should also get used to specify version requirements in our package.xml to make these explicit, like this : https://github.com/asmodehn/pyros/blob/config_refactor/package.xml

BTW : If you have a look at celeros and rostful you will see that I use submodules for dependencies : 
https://github.com/asmodehn/rostful/tree/indigo-devel/deps

I will, in the next few weeks, simplify these submodule dependencies and instead make ros packages for all the dependencies, using https://github.com/asmodehn/catkin_pure_python : it makes creating catkin packages from recent (setuptools) python packages much easier ( no need to modify setup.py, only CMakeLists.txt and package.xml are needed ).
It is still in development though, but feel free to try it, and contact me if anything causes you trouble.

asmodehn:
I had a look at tornado versions on ubuntu.

[...]

PR #317, opened 2018-01-11
Removed tornado source code, added dependency

MBlistein:
There are debian packages for both tornado and backports, that contain newer versions as the source code copied inside of rosbridge_server. 
Therefore the source code has been deleted and replaced with run dependencies on the corresponding debian packages.


T045T:
There have been attempts to switch to an external version before (see #162 for an overview and links to related issues).

I'm not sure if there aren't any subtle issues with using an external `tornado`.
@rctoris , @jihoonl , @jonbinney: Looks like you were involved in the previous effort, do you know of any showstoppers here?

rosbridge_server/package.xml
```diff
@@ -15,12 +15,10 @@
...
-  <build_depend>rosbridge_library</build_depend>
-  <build_depend>rosapi</build_depend>
-  <build_depend>rospy</build_depend>
+  <run_depend>python-backports.ssl-match-hostname</run_depend>
+  <run_depend>python-tornado</run_depend>
   <run_depend>python-twisted-core</run_depend>
   <run_depend>rosbridge_library</run_depend>
   <run_depend>rosapi</run_depend>
   <run_depend>rospy</run_depend>
-  <run_depend>rosauth</run_depend>
```
jihoonl: don't remove `rosauth`. It is in the code

MBlistein: Ok, @jspricke will create a debian package and we will look at testing together with @Behery  next week.

MBlistein:
Hey,
big thanks to jspricke who made a debian package and Behery who tested it. It seems to work out:
- when installing / upgrading, it overwrites the existing installation in opt/ros/kinetic
- tornado and backport are uninstalled (but corresponding debian packages installed via apt)
- the rosbridge server is functional:
- [ ]  subscribing/ unsubscribing/ publishing to topics works
- [ ]  service calls go through, advertising a service also works

Should be good to go, I will update the pull request with your remark in a minute.

jihoonl:
Since we plan to squash and merge, it should be fine to have a merge commit. It will get squashed as long as there is no conflict.

jihoonl:
Let me ping ros maintainer to have a last check.

MBlistein:
Ok, thanks! I did rebase in the meantime.

jihoonl:
@MBlistein hey I just released the newer version of rosbridge. You may want to test `ros-shadow-fixed` for final checking.

MBlistein:
@jihoonl Sorry for the late reply. Tested it (thanks to @Behery ) and it seems fine. Luckily, since already released out of shadow-fixed :)

PR #319, opened 2018-01-19
message_conversion: create stand-alone object inst

onjen:
Catching the ROSInitException allows to create object
instances without an initialized ROS state


rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -275,8 +275,11 @@ def _to_object_inst(msg, rostype, roottype, inst, stack):
         raise FieldTypeMismatchException(roottype, stack, rostype, type(msg))
 
     # Substitute the correct time if we're an std_msgs/Header
-    if rostype in ros_header_types:
-        inst.stamp = rospy.get_rostime()
+    try:
+        if rostype in ros_header_types:
+            inst.stamp = rospy.get_rostime()
+    except rospy.exceptions.ROSInitException as e:
+        rospy.logdebug("Not substituting the correct header time: %s" % e)
```
jihoonl: `rospy.logdebug` will also fail because ros logger only works after `rospy.init_node`. 

Can you provide me some context why you need this?

T045T: The logging functions actually don't raise any exceptions if you call them before `init_node`, so this should be fine.

onjen: This will be used to create message objects from offline data into a rosbag.

The logging function doesn't raise exceptions, it runs fine like this.

PR #330, opened 2018-05-09
removed symbolic links from scripts directory and changed launch file to start python scripts

ozfaro:
The launch files refer to the symbolic links in the scripts directory of rsobridge_server (e.g. rosbridge_websocket), not rosbridge_websocket.py.
The symbolic links in the scripts folder can cause problems when a dpkg package is generated from the package. As a result the files in the dpkg package (e.g. rosbridge_websocket) only contain the text rosbridge_websocket.py. In the case of rosbridge_tcp the link is set to ./rosbridge_tcp what results in a file that contains ./rosbridge_tcp and bascially works as a link. Unfortunately this only works for ./rosbridge_tcp not for rosbridge_websocket and rosbridge_udp, because here the links are set directly to rosbridge_websocket, etc.

solution: remove the symbolic links and call the py-files directly in the launch file

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/publishers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/publishers.py
@@ -310,7 +310,7 @@ class PublisherManager():
         if not topic in self._publishers:
             return
 
-        self._publishers[topic].unregister_client(client_id)
+        #self._publishers[topic].unregister_client(client_id)
         if topic in self.unregister_timers:
             self.unregister_timers[topic].cancel()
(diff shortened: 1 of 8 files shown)


jihoonl:
Looks good to me to change roslaunch files to use .py directly. But, I would prefer not to remove the symlink files for backward compatibility reason. Can you put it back?

ozfaro:
@jihoonl: yes sounds reasonable. Will create a new reuqest with updated launch files

mvollrath:
Looks like a new PR was created?  Closing this one.

PR #337, opened 2018-07-06
removing has_key for python3, keeping backwards compatibility

klintan:
Pretty straight forward, remove has_key method for dict since it is removed in python 3. Changes are still backwards compatible with python 2.x.


klintan:
Added an additional change for py3 (but keeping py2 compatibility). 

I don't think it should affect performance, because it's basically just a hash-lookup instead of a generator, but it might I guess?

jihoonl:
returning list instead of iterator but it should be fine. merging.

PR #364, opened 2018-11-04
CBOR encoding

mvollrath:
Add "cbor" compression type, which produces binary messages encoded with [CBOR](https://tools.ietf.org/html/rfc7049).  For large messages with stochastic data like point clouds, this is more efficient than JSON or PNG encodings.  Decoding byte arrays is very fast in Chrome.

[Draft typed array tags](https://tools.ietf.org/html/draft-ietf-cbor-array-tags-00) are implemented for fast decoding of homogeneous numeric arrays.

A roslibjs patch is on the way.


mvollrath:
Switched to rosdepped CBOR library, now depends on https://github.com/ros/rosdistro/pull/19377

mvollrath:
Did some more testing and found that CBOR is fastest for large numeric arrays.  64-bit integers may be an exception because JavaScript does not support them natively.  PNG can result in a smaller wire size but at great expense on both ends.  Per-message deflate at the websocket level is a better way to reduce wire size for browser clients.

achim-k:
@mvollrath the rosdistro PR was merged. Can you re-trigger the Travis build? Not sure how to do this though, maybe closing and re-opening the PR will do

PR #365, opened 2018-11-06
Add param to enable ws per-message deflate

mvollrath:
Tornado has its own per-message deflate compression option, which
compresses each WebSocket message.  The compression level should be
roughly equivalent to PNG compression, depending on whether the message is
JSON or binary (CBOR).  The encoding/decoding time will be much faster
than protocol PNG compression.

This param should be enabled when wire size is important, e.g. not
connecting to localhost.

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -47,6 +47,7 @@ class RosbridgeWebSocket(WebSocketHandler):
     clients_connected = 0
     client_count_pub = None
     authenticate = False
+    use_compression = False
 
     # The following are passed on to RosbridgeProtocol
     # defragmentation.py:
@@ -126,3 +127,11 @@ class RosbridgeWebSocket(WebSocketHandler):
 
     def check_origin(self, origin):
         return True
+
+    def get_compression_options(self):
+        cls = self.__class__
+
+        if not cls.use_compression:
+            return None
+
+        return {}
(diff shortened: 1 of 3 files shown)


rosbridge_server/src/rosbridge_server/websocket_handler.py
```diff
@@ -126,3 +127,11 @@ def send_message(self, message):
...
     def check_origin(self, origin):
         return True
+
+    def get_compression_options(self):
+        cls = self.__class__
+
+        if not cls.use_compression:
+            return None
+
+        return {}
```
viktorku: Do we want to document the behaviour of this? At least a copy paste from [Tornado's docs](https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler.get_compression_options) can help:

> If this method returns None (the default), compression will be disabled. If it returns a dict (even an empty one), it will be enabled. 

Should we expose the [compression and memory](https://docs.python.org/3.6/library/zlib.html#zlib.compressobj) level as well? IMO at least a comment describing the defaults would suffice.

mvollrath: I didn't see any of the websocket server's params documented anywhere, I'm happy to add a note if you can point me at a place to leave it.  I assume the Tornado docs will be the first point of reference for anybody looking at this.

Compression and memory level only became available in Tornado 4.5, so if we exposed them it might take extra work to support kinetic.  I didn't feel like adjusting those settings would be particularly valuable.

PR #385, opened 2019-01-31
Synchronous websocket write

### LINKED ISSUE #212: Memory exhaustion on server with high rate large blob data
This is follow-up issue of #192.

_Questions_
- Should rosbridge have intelligence or configuration to adapt slow and fast clients seamlessly?
- Rosbridge server will suffer from memory exhaustion if slow client is subscribing to high rate large chuck data anyway. How should rosbridge react on this memory exhaustion?

mvollrath:
Fixes #212 by blocking `send_message()` while a message is being written.

In practice this connects the backpressure to either the `rospy.Subscriber` `Thread` (in the case of `MessageHandler` or `ThrottleMessageHandler`) or the `QueueMessageHandler` `Thread` in that case.

Applications should subscribe with `queue_length` of at least 1 to prevent pushing all the way back to the `rospy.Subscriber`, which is probably rude to other clients, and/or use the ["workersocket" transportLibrary](https://github.com/RobotWebTools/roslibjs/pull/317) to decouple the `WebSocket` from the client main thread.

There's still a way to DoS other clients (0 queue_length, never ACK) but this is a net positive for the websocket server:

* `QueueMessageHandler` actually works instead of dumping messages into Tornado's queue immediately.
* Server doesn't OOM when the client can't keep up with incoming data.

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -34,10 +34,12 @@ import rospy
 
 from rosauth.srv import Authentication
 
+import threading
 from functools import partial
 
 from tornado.ioloop import IOLoop
 from tornado.websocket import WebSocketHandler
+from tornado.gen import coroutine
 
 from rosbridge_library.rosbridge_protocol import RosbridgeProtocol
 from rosbridge_library.util import json, bson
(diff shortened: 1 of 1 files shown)


jihoonl:
It looks good. Will this work with tornado 4.2.1(the default tornado version in xenial)?

PR #391, opened 2019-03-14
BSON can handle NaN,Inf , null workaround is not needed

akirayou:
In case of bson_only_mode, you don't have to use "null workaround" for JSON.
It's useful on C# (JSON.NET). If you want use float with null in C# , you have to use "float?" type.
It's slow and you have to cast to "float" type, every time .

It's JSON's restriction, there is no restriction on BSON,


jihoonl:
:+1: thanks

PR #395, opened 2019-03-27
Handle BadYieldError in Tornado <4.5.0

### LINKED ISSUE #394: BadYieldError log spamming
Just updated to `rosbridge_server` v0.10.2 and I'm now being spammed with the following log message hundreds of times per second
```
[ERROR] [1553705564.030598]: Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rosbridge_server/websocket_handler.py", line 168, in prewrite_message
    yield self.write_message(message, binary)
  File "/usr/local/lib/python2.7/dist-packages/tornado/gen.py", line 628, in run
    value = future.result()
  File "/usr/local/lib/python2.7/dist-packages/tornado/concurrent.py", line 111, in result
    raise self._exception
BadYieldError: yielded unknown object None
```

As far as I can tell, the immediate cause was the addition of exception logging in #386, but it looks like this exception was being thrown before then. As far as I can tell this exception has no actual impact on the functionality of the server, but after #386 produces a lot of noise.

mvollrath:
Fixes #394

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -39,9 +39,10 @@ import threading
 import traceback
 from functools import partial, wraps
 
+from tornado import version_info as tornado_version_info
 from tornado.ioloop import IOLoop
 from tornado.websocket import WebSocketHandler, WebSocketClosedError
-from tornado.gen import coroutine
+from tornado.gen import coroutine, BadYieldError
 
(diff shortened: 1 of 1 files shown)


mvollrath:
The exception started being raised due to #385 but wasn't logged until #386 , this is probably why I didn't see it in my testing of the individual branches on different distros.  You find all kinds of things when you actually log errors!

PR #408, opened 2019-06-08
Fixes #403 : yield infinite lock

### LINKED ISSUE #403: rosbridge_websocket: WebSocket opening handshake timed out
## Expected Behavior
`rosbridge_websocket` should never stops to accept new connections

## Actual Behavior
We use `rosbridge_websocket` with our web app (via roslibjs). Every time we stat the bridge and our app, everything works as expected. However after a random amount of time, `rosbridge_websocket` stop to accept new connections and we get the following error (on chrome web console):
```
WebSocket connection to 'ws://localhost:9090/' failed: WebSocket opening handshake timed out
```
Once the rosbridge is in this state, we've tried to connect from other browser or even from a bare websocket client and we have no response from the bridge.


If we check the logs of the bridge, without the problem, we have some interesting data:

```
[rospy.internal][INFO] 2019-04-25 15:39:44,522: topic impl's ref count is zero, deleting topic /xxxx...
[rospy.internal][INFO] 2019-04-25 15:39:44,522: topic[/xxxx] removing connection to /yyyy
```
But we never get a 
```
[rosout][INFO] 2019-04-25 15:39:01,735: Client disconnected. 0 clients total.
```
As usual.

Please note that our web app automatically reconnect when connection is lost. That may lead to a very fast  disconnection/reconnection.

### More data that could help you to troubleshoot:

The bridge seems to think that there is still a client connected:

```
rostopic echo /connected_clients          
clients: 
  - 
    ip_address: "127.0.0.1"
    connection_time: 
      secs: 1556199542
      nsecs: 117264986
---
```
```
rostopic echo /client_count                                                                                                                                                                                                                                                     130 ↵
data: 1
---
```

The bridge is still registred as listening on the linux network stack:
```
ss -lp | grep 9090
tcp               LISTEN              9                    128                                                                                          0.0.0.0:9090                                                0.0.0.0:*                    users:(("python",pid=14397,fd=13))                                             
tcp               LISTEN              10                   128                                                                                             [::]:9090                                                   [::]:*                    users:(("python",pid=14397,fd=14))  
```



## Specifications

  - ROS Version: melodic
  - OS Version: bionic
  - Rosbridge Version: `0.11.0` (we've just updated, but the issue is not new)
  - Tornado Version: 4.5.3

Alabate:
More on #403

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -167,7 +167,8 @@ class RosbridgeWebSocket(WebSocketHandler):
         # Use a try block because the log decorator doesn't cooperate with @coroutine.
         try:
             with self._write_lock:
-                yield self.write_message(message, binary)
+                future_handle = self.write_message(message, binary)
+            yield future_handle
         except WebSocketClosedError:
             rospy.logwarn('WebSocketClosedError: Tried to write to a closed websocket')
             raise


jihoonl:
@mvollrath  could you take a look?

mvollrath:
It looks like this would defeat the behavior of waiting for a write to finish before sending another write to Tornado, effectively disabling all of rosbridge queueing features and allowing clients to crash the server by subscribing to high rate topics and never ACKing messages.

Alabate:
Thank you for your answer. I didn't understood that. So if moving the yield outside of the lock is not an option, we need to find why at some point the generator is not called anymore and freezing the full server not even listening to sigint.

I'm closing this pr, but if you could take a look at #403, that would be really nice. Thanks!

PR #410, opened 2019-06-09
Add settings for websocket timeout

Alabate:
I've noticed that if the connection between websocket client and server is closed unexpectedly, a client can be marked as connected forever on the server. With only one client I managed to make the `client_count` topic grow higher than 3 by disabling wifi while rosbridge was connected.

Tornado has parameters to enable the keep-alive ping which is integrated into [websocket protocol](https://tools.ietf.org/html/rfc6455#section-5.5.2). I exposed `websocket_ping_interval` and `websocket_ping_timeout` as ROS and CLI parameters. [Tornado documentation of those parameters](https://www.tornadoweb.org/en/stable/web.html#tornado.web.Application.settings).

I've set `websocket_ping_interval` to 0, to let this feature disabled by default. Because if someone use rosbridge with a buggy websocket client, it might not answer to the ping. So, it can be a breaking change in some case if we enable it by default. I let you decide if you want to enable it by default.

### DIFF
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -79,6 +79,11 @@ if __name__ == "__main__":
     if RosbridgeWebSocket.max_message_size == "None":
         RosbridgeWebSocket.max_message_size = None
 
+    # get tornado application parameters
+    tornado_settings = {}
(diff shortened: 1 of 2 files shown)


Alabate:
I've added error handling of `StreamClosedError` because this is an expected error when client is closing because of timeout and there a still stuff scheduled to be sent to him. Also, don't need to have the information for each message, just once every second is enough to understand what happened.

PR #423, opened 2019-07-31
Fixes #418: WebSocketClosedError Spam

### LINKED ISSUE #418: WebSocketClosedError: Tried to write to a closed websocket
We upgraded yesterday to rosbridge_suite 0.11.2 and on the webserver side we get now an infinite-repeating of the following warning: 

WebSocketClosedError: Tried to write to a closed websocket 

However, our webapp is still running and roslib.js is not complaining on the client side. I just wonder if that error refers to unsubscribed topics or users that disconnected.

## Specifications

  - ROS Version (`echo $ROS_DISTRO`): kinetic
  - OS Version (`grep DISTRIB_CODENAME /etc/lsb-release`): xenial
  - Rosbridge Version (`roscat rosbridge_server package.xml | grep '<version>'`): 0.11.2
  - Tornado Version (`python -c 'import tornado; print tornado.version'`): 4.2.1

lennartdopatka:
#413 added a check for the result of tornados WebSocketHandler.write_message and is throwing WebSocketClosedError if its None.
WebSocketHandler.write_message does return a Future since version 4.3.0. ([tornado write_message doc](https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler.write_message))
I added a check to only raisie the Error when version is >4.3.0, to prevent anyone using a tornado version <4.3.0 to have a WebSockedClosedError for every message (see #418).

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -172,6 +172,7 @@ class RosbridgeWebSocket(WebSocketHandler):
 
                 # When closing, self.write_message() return None even if it's an undocument output.
                 # Consider it as WebSocketClosedError
+                # For tornado versions <4.3.0 self.write_message() does not have a return value
                 if future is None and tornado_version_info >= (4,3,0,0):
                     raise WebSocketClosedError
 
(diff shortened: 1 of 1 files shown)


mvollrath:
Thanks you!

PR #428, opened 2019-08-14
Add compression to service responses

uahic:
TODOS

- Add tests for service call with compression flags 'png' and 'cbor'
- The arguments of the service calls are currently (in my understanding) not compressed, I will have a look at this soon


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
mvollrath: This file shouldn't be in the repo, it may interfere with other developers.  The .vscode path should really be in .gitignore, but you don't have to do that.

mvollrath:
@uahic still working on this?

uahic:
@mvollrath Unfortunately I lack the time right now due to a bunch of new projects :- (

PR #433, opened 2019-09-19
Migrate launch files to ROS 2

hidmic:
This pull request migrates all `roslaunch` XML files in the `rosbridge_server` package to ROS 2 launch XML. Note though that currently only the websocket handler is supported by the implementation.


dirk-thomas:
Can you maybe also update the launch-based tests?

hidmic:
I can update those tests, but I'd rather do so in another PR as it's not just a change of testing framework. Tests are currently disabled and thus still bound to ROS 1:

https://github.com/RobotWebTools/rosbridge_suite/blob/ba79f3504bbea1f075e13c1d77b9293829d9abcd/rosbridge_library/CMakeLists.txt#L42-L47

PR #436, opened 2019-09-24
Update rosbridge_suite packages' dependencies.

hidmic:
Precisely what the title says. It also removes an unused dependency.


dirk-thomas:
Thanks for the fix.

PR #447, opened 2019-11-26
Update rosapi/proxy.py to match eloquent API

justinscorringe:
Updates import to match eloquent update

Update get_service_info to get_service_server_info


dirk-thomas:
@justinscorringe The usage needs to be updated too.

dirk-thomas:
Thanks for the patch.

PR #449, opened 2019-12-05
Install scripts with catkin_install_python.

mikepurvis:
This is friendlier to using alternative pythons, and the recommended approach:

http://docs.ros.org/melodic/api/catkin/html/howto/format2/installing_python.html#scripts


mikepurvis:
Looks like the test failure is a flaky `100 != 101` thing and can be ignored/retriggered.

PR #455, opened 2020-02-07
FieldTypeMismatchException error in python3

TeamATR:
type bool changed to boolean in python3

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -47,7 +47,7 @@ from rosbridge_library.util import string_types, bson
 import sys
 if sys.version_info >= (3, 0):
     type_map = {
-    "bool":    ["bool"],
+    "bool":    ["boolean"],
     "int":     ["int8", "byte", "uint8", "char",
                 "int16", "uint16", "int32", "uint32",
                 "int64", "uint64", "float32", "float64"],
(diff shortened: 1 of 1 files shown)


mvollrath:
Does this work in Python 2?

How can we reproduce the original issue?

TeamATR:
@mvollrath 
I did not change anything which will affect while using Python2.
1. `type_map` is different due to the Python version.
2. `ros_primitive_types` is used for both Python versions, so I added `boolean` for Python 3 instead of changing `bool` to `boolean` so might affect Python 2.
3. Probably, should define a different `ros_primitive_types` according to the Python version which uses `boolean` just like `type_map`.

To reproduce the original issue, without changes, using Python 3, call `std_srvs/srv/SetBool` type service and pass data `{"data": True}`, will cause FieldTypeMismatchException.
e.g.
```python
service = roslibpy.Service(client, '/set_ludicrous_speed', 'std_srvs/srv/SetBool')
request = roslibpy.ServiceRequest({"data": True})
result = service.call(request)
```

mvollrath:
I can't reproduce this issue in melodic, maybe there is some other environmental condition I'm missing?

TeamATR:
I am sorry. I forgot to mention. I used it for ROS2-dashing env

jihoonl:
I can't find any information that saying python3 uses boolean instead of bool. Would you provide where you found the change?

TeamATR:
@jihoonl  @mvollrath Apologize! It is not because of Python version. It is the class `std_srvs.srv._set_bool.py` in my environment using defines `boolean` instead of using`bool`. 

In this case, where should I report the problem? Or it is just my local multiple libs compatibility problem?

jihoonl:
There is `ros2` branch which you may be interested in. Would you try ros2 branch?

TeamATR:
Thank you! Sure! I will close this pull request,  make a pull request to `ros2` branch.

PR #463, opened 2020-02-24
Support Local file origins

Axel13fr:
Tornado wasn't making any checks on request origins but Autobahn does. 
So we "lost" a feature when switching to autobahn: allowing websocket connection when browsing from a local file.

Allowing null origin optionally to support browser local file origins: file:// for Chrome or empty origins for Firefox are treated as null origins by Autobahn. But null origins are disabled by default.


mvollrath:
`externalPort` was added in Autobahn 0.15.0, you can compare to `autobahn.version` to gate this feature for kinetic.

mvollrath:
See https://github.com/RobotWebTools/rosbridge_suite/pull/469 for a portable way to version gate this feature.

Axel13fr:
Duplicate

PR #464, opened 2020-03-05
Fix queue blocking

mvollrath:
`QueueMessageHandler.run` was holding onto its lock while writing out to the server.  Since the websocket server can block outgoing messages now, this was causing cross-client blocking by jamming up the rospy Subscriber thread.

By releasing the lock before writing, we let this handler work the way it was always intended.

The handler's internal queue was switched to a `deque` for simplicity.


reinzor:
We will test with this the coming weeks and see if the `hanging` still occurs.

reinzor:
Still seems to occur, some topics are not coming in anymore. Latched topics keep old state and are not updating.

mvollrath:
@jihoonl @viktorku any questions?

mvollrath:
@jihoonl I've added a test which fails on develop to illustrate what this is doing.

jihoonl:
Sorry for late response. Looks good. I have full trust on you. thank @mvollrath!

PR #468, opened 2020-03-11
Expose autobahn externalPort parameter

### LINKED ISSUE #467: Autobahn externalPort setting should be exposed
Running rosbridge_websocket behind a reverse proxy, for instance nginx, results in a connection drop during the handshake due to a different port specified in the original request header. Here the error message:

```
[-] failing WebSocket opening handshake ('port 6587 in HTTP Host header '127.0.0.1:6587' does not match server listening port 9090')
[-] dropping connection to peer tcp4:127.0.0.1:60550 with abort=False: port 6587 in HTTP Host header '127.0.0.1:6587' does not match server listening port 9090
```

These issues opened in the autobahn github repository also describe the issue:

https://github.com/crossbario/autobahn-python/issues/775
https://github.com/crossbario/autobahn-python/issues/495

Such parameter `externalPort` should be passed to the WebSocketServerFactory constructor:

https://github.com/RobotWebTools/rosbridge_suite/blob/5a0efdcb785fe354780cd2b6897b17809892bc84/rosbridge_server/scripts/rosbridge_websocket.py#L273

## Expected Behavior

Connection should be established behind a reverse proxy and a different port

## Actual Behavior

Connection drops

  1. Run rosbridge_websocket on port 9090
  2. Configure a reverse-proxy in front of rosbrige on a different port and try to connect through it.

## Specifications

  - ROS Version: melodic
  - OS Version: bionic
  - Rosbridge Version: 0.11.4
  - Twisted Version: 17.9.0

epaezrubio:
Fixes #467

### DIFF
--- a/rosbridge_server/launch/rosbridge_websocket.launch
+++ b/rosbridge_server/launch/rosbridge_websocket.launch
@@ -16,6 +16,10 @@
 
   <arg name="authenticate" default="false" />
 
+  <arg name="websocket_ping_interval" default="0" />
+  <arg name="websocket_ping_timeout" default="30" />
+  <arg name="websocket_external_port" default="None" />
+
   <arg name="topics_glob" default="[*]" />
   <arg name="services_glob" default="[*]" />
   <arg name="params_glob" default="[*]" />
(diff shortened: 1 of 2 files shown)


epaezrubio:
I've followed the same approach as for `websocket_ping_interval` and `websocket_ping_interval`. I was surprised not to find them here in the respective launch file https://github.com/RobotWebTools/rosbridge_suite/blob/develop/rosbridge_server/launch/rosbridge_websocket.launch , but still I didn't add `websocket_external_port` to the launch file either.

Is it necessary to create a second PR adding all of them?

mvollrath:
You can add the params in this PR if you like.

ejalaa12:
Sorry for double commenting:
https://github.com/RobotWebTools/rosbridge_suite/commit/a8179375ef0545226cee616907f64b292e9dbc29#r38031376

But as I said, the last commit breaks the launch file as the arg won't be correctly interpreted in the python node. I guess the arg/param `websocket_external_port` must be simply removed from the launch.

Or we could replace the default value like this:
```xml

  <arg name="websocket_external_port" default="$(arg port)" />
```

epaezrubio:
Good point, just tested it and was able to reproduce the error 👍

PR #469, opened 2020-03-14
Default to supporting local files as we had before Autobahn.

boilerbots:
For Kinetic 0.10.3 version don't add the allowNullOrigin option.
For everything else add it and default to True because that is
consistent with the original behaviour of this package.

### DIFF
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -38,6 +38,7 @@ import sys
 from twisted.python import log
 from twisted.internet import reactor, ssl
 from twisted.internet.error import CannotListenError, ReactorNotRunning
+import autobahn #to check version
 from autobahn.twisted.websocket import WebSocketServerFactory, listenWS
 from autobahn.websocket.compress import (PerMessageDeflateOffer,
                                          PerMessageDeflateOfferAccept)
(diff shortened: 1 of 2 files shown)


rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -272,11 +274,19 @@ def handle_compression_offers(offers):
     uri = '{}://{}:{}'.format(protocol, address, port)
     factory = WebSocketServerFactory(uri)
     factory.protocol = RosbridgeWebSocket
-    factory.setProtocolOptions(
-        perMessageCompressionAccept=handle_compression_offers,
-        autoPingInterval=ping_interval,
-        autoPingTimeout=ping_timeout,
-    )
+    if autobahn.__version__ == '0.10.3':
```
mvollrath: Comparing to a specific version is very brittle and not likely to work for users of other OS distros.  Since the distutils package is expected to be available in ROS distributions and we know exactly which version adds the feature, please use [`distutils.version.LooseVersion`](http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html).  e.g.
```python
from distutils.version import LooseVersion

...

if LooseVersion(autobahn.__version__) < LooseVersion("0.15.0"):
    ...
```

boilerbots: good idea, if you know a better way and are a maintainer can you fix this bug? Otherwise I will submit another fix when I can.

mvollrath:
@jihoonl can you trigger a retest?

mvollrath:
Nevermind, found a place to edit.

mvollrath:
Merge pending #473

PR #478, opened 2020-04-07
Subscriber concurrency review

mvollrath:
Protect critical sections in SubscriberManager which were previously unprotected from race conditions between clients.

Raise Exception when unsubscribe input is invalid instead of passing this condition silently.


jihoonl:
:+1:

PR #493, opened 2020-04-28
ros2 - Adaptaions to Eloquent

mmatthe:
During usage of the ROS2 branch of the rosbridge_suite, several problems occured in the Python Code. These were fixed such that it works for our use case. The overview of the changes:

    allow an increased subscribing frequency of 1000Hz (it was 10Hz, meaning only 10 ROS2 messages were received per second)
    Speed up receiving CompressedImage from the websocket by creating a shortcut implementation with some optimizations
    allow both nsecs and nanosec fields for time (ROS1/2 compatiblity).
    allow received int values to be interpreted as float values, when needed.


mvollrath:
I'm happy to merge this with a review by a ROS2 user.

flynneva:
Tested and confirmed on Ubuntu 18.04 with ROS2 Eloquent.

**NOTE:** I had to `pip install pymongo` before using `rosbridge_websocket`. 

tested with webviz.io resulted in a service call error. I think the service that webviz was looking for just was missing.

tested on a less-complex website and it worked great. 

pymongo_dep

flynneva:
just tested it with dashing too...works great

mvollrath:
Please try removing the pip-installed pymongo package and test with apt package `python3-bson`, because this is what will be installed for BSON library.

It looks like the webviz error would go away if the rosapi node (part of rosbridge) were running.

flynneva:
followed your recommendation - uninstalled the pip pymongo and installed the apt package python3-bson. 

same as before, compiled and worked great.

I also tried to launch the rosapi node in another terminal, however it looks like the service `/rosapi/topics_and_raw_types` that webviz is looking for isnt available. it probably wasn't carried over to the ros2 version of rosapi yet.

rosapi_service_list

mvollrath:
Is there something new in ROS2 where having a requirements.txt is either supported or desired?

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -133,15 +137,23 @@ def extract_values(inst):
...
+    inst.header.stamp.nanosec = msg["header"]["stamp"]["nanosec"]
+    inst.format = msg["format"]
+    inst.data = array.array("B", msg["data"])
+    return inst
 
 def populate_instance(msg, inst):
     """ Returns an instance of the provided class, with its fields populated
     according to the values in msg """
     inst_type = msg_instance_type_repr(inst)
+    if inst_type == "sensor_msgs/CompressedImage":
```
mvollrath, line 152: I'm not convinced yet that hardcoding special handling for specific message types in this module in this way is a good idea.

If the data is coming in as anything but a base64 string, it shouldn't be coming anywhere near this code path.  I don't see any place where base64 decoding was added, so I don't understand how this decoder could possibly work.  Can you explain how this works?

requirements.txt
```diff
@@ -0,0 +1,12 @@
+colcon-ros==0.3.14
```
mvollrath, line 1: Can you show me where in the ROS2 docs it says that requirements.txt is supported or recommended in ROS2 as an alternative to rosdep?

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -34,6 +34,10 @@
 from __future__ import print_function
 import rclpy
 from rclpy.clock import ROSClock
+import numpy as np
+import array
+import time
+import random
```
mvollrath, line 40: This import is not used.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -34,6 +34,10 @@
 from __future__ import print_function
 import rclpy
 from rclpy.clock import ROSClock
+import numpy as np
+import array
+import time
```
mvollrath, line 39: This import is not used.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -233,6 +249,7 @@ def _to_inst(msg, rostype, roottype, inst=None, stack=[]):
         return _to_primitive_inst(msg, rostype, roottype, stack)
 
     # Check whether we're dealing with a list type
+    #print ("Inst: ", inst, type(inst))
```
mvollrath, line 252: Please clean up these commented-out print statements.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -292,14 +315,19 @@ def _to_primitive_inst(msg, rostype, roottype, stack):
 
 
 def _to_list_inst(msg, rostype, roottype, inst, stack):
+    #print ("To list", msg, type(msg), rostype, roottype, inst, stack)
     # Typecheck the msg
     if type(msg) not in list_types:
         raise FieldTypeMismatchException(roottype, stack, rostype, type(msg))
 
+
```
mvollrath, line 323: Please clean up extreneous whitespace.

rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -105,6 +107,7 @@ def __init__(self, topic, client_id, callback, node_handle, msg_type=None):
         self.msg_class = msg_class
         self.node_handle = node_handle
         # TODO(@jubeira): add support for other QoS.
+        print ("Subscriber in __init__: ", msg_class, "topic: ", topic, "cb: ", self.callback)
```
mvollrath, line 110: Please remove the extreneous `print` logging in this module.  Use `logging` module debug level if you really want this level of verbosity.

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -118,6 +118,9 @@ def incoming(self, message_string=""):
         message_string -- the wire-level message sent by the client
 
         """
+        if type(message_string) is bytes:
```
mvollrath, line 121: This decoding should be done at the rosbridge_server level before it makes it to this point, unless you are implementing support for incoming binary messages in this PR, which does not appear to be the case.

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -202,6 +205,9 @@ def incoming(self, message_string=""):
         try:
             self.operations[op](msg)
         except Exception as exc:
+            # import traceback as tb
```
mvollrath, line 208: Please clean up commented-out print statements in this module too.

requirements.txt
mmatthe: Well, officially `rosdep` shall be used to manage requirements and dependencies, which are described in `package.xml`. However, I personally hate don't like when some tooling changes my base system (rosdep uses apt-get to install dependencies). I prefer to have my source isolated. That's why I provide this requirements.txt file, whilch allows me to run rosbridge within a venv. 

However, if you prefer, I can delete the file from this repo. 

Please let me know your preference.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
mmatthe: I agree that this is kind of a hacky solution to shortcut decoding of `CompressedImage`. At this point in the data flow, msg is a `dict` containing all the fields. In particular, arrays/lists have already been decoded into python/numpy arrays. From this point in the data flow on, the only remaining thing is to translate the fields from the dict into the attributes of `inst` (`inst` is already the target object of the correct type). in our debugging sessions it happened that decoding of CompressedImages was extremely slow. This is due to [an known problem](https://github.com/ros2/rosidl_python/issues/79) within the python-idl mapping. Particular, `inst.data = msg["data"]` would be the slow thing. Instead, it should be `inst.data = array.array("B", msg["data"])` which omits the runtime check if all values are within bounds. 

Since the generic decoding within `populate_instance` can only do `inst.data = msg["data"]`, I have included the shortcut here. 

Please let me know, if I should delete it in this fork or keep it.

rosbridge_library/src/rosbridge_library/protocol.py
mmatthe: I have moved it into `websocket_handler.py`

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
mvollrath: It seems like this could be made into a general solution by moving it into `_to_inst` with the rest of the decoding.  Otherwise I would accept this hack in this PR if there is a comment linking to [your comment explaining why it exists](https://github.com/RobotWebTools/rosbridge_suite/pull/493#discussion_r425151919).

requirements.txt
mvollrath: The main reason I don't want requirements.txt in the repo is because it becomes everybody else's job to maintain your requirements.txt.  We already have one list of dependencies to maintain.

mmatthe: I have deleted the requirements.txt file from this fork.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
mmatthe: I have created an extra file that collects shortcut decoders. Currently, it only contains a decoder for `sensor_msgs/CompressedImage`. I needed that file to keep the `message_conversion.py` clean.  I have moved the call to these function into the `_to_object_inst`

Please see, if this solution is fine with you.

flynneva:
just tested this for foxy on ubuntu 20.04 - works there too!

maxlein:
I am confused on which ros2 bridge I shall use now!?  
This one on ros2 branch or [ros2-web-bridge](https://github.com/RobotWebTools/ros2-web-bridge)?  
Seems both are missing `topics_and_raw_types` service for usage with webviz.  
Am I missing something? Are they dependent on each other somehow?

flynneva:
either one! this ones based on python and the ROS2-bridge one is based on js. they provide similar functionality

mvollrath:
> Seems both are missing `topics_and_raw_types` service for usage with webviz.

IIRC this service is provided by `rosapi` which is a separate script under rosbridge_suite.  I don't know if rosapi has been ported for ROS2 yet.

rosbridge_library/src/rosbridge_library/internal/object_decoders.py
```diff
@@ -0,0 +1,53 @@
...
+def decode_compressed_image(msg, inst):
+    inst.header.frame_id = msg["header"]["frame_id"]
+    inst.header.stamp.sec = msg["header"]["stamp"]["sec"]
+    inst.header.stamp.nanosec = msg["header"]["stamp"]["nanosec"]
+    inst.format = msg["format"]
+
+    # explicitely cast to array.array to overcome a costly runtime
+    # check within inst.data instance.
+    # See also https://github.com/RobotWebTools/rosbridge_suite/pull/493#discussion_r425151919
+    inst.data = array.array("B", msg["data"])
```
mvollrath, line 49: Is it not true that if you did this for all `uint8[]` in message_conversion.py `_to_binary_inst` you wouldn't need special cases for specific message types?

mmatthe: Well, you can, in the deeply buried code where the instance is set (it'd be in `_to_object_inst` and `_to_list_inst` capture if a uint8[] is set and then apply the optimization. However, this is not very clear to understand and it might be too general, as sometimes you might want to have the additional error check that is done by ROS2 idl_typesupport. Therefore, I would not recommend using this optimization generally.

mvollrath: I don't see how maintaining unique decoder logic for specific message types is better.  Everything about rosbridge is general and should remain general.  If the usual decoding pathway is too slow for CompressedImage it's probably too slow for every other message that uses `uint8[]`.  If the additional error check can be bypassed for CompressedImage it can probably be bypassed for every other message type.

If there is an open issue about ROS2 message conversion performance somewhere else please link.

travipross:
Also compiled and tested on Ubuntu 18.04.3 with ROS Eloquent. Seems to work :+1:

AviKenz:
> Tested and confirmed on Ubuntu 18.04 with ROS2 Eloquent.
> Compiled, no issues.
> 
> compiled
> 
> **NOTE:** to run, I had to `pip install pymongo` before using `rosbridge_websocket`.
> 
> tested with webviz.io resulted in a service call error. I think the service that webviz was looking for just was missing.
> 
> tested on a less-complex website and it worked great.
> 
> pymongo_dep

Hello !
i am getting similar output and i CANT publish anything from any Client library.
i already tried roslibjs and jrosbridge. same errors...

Here is the issue i opened: https://github.com/rctoris/jrosbridge/issues/29

I need help pleaaaaase....

im running ros-dashing.

Thanks in advance !

AviKenz:
> just tested it with dashing too...works great

im using dashing too. but nothing works. 

my issue: https://github.com/rctoris/jrosbridge/issues/29

need help !

flynneva:
@mmatthe if you make one more small commit on your PR branch it should trigger the new ROS2 CI for this PR.  might be helpful in diagnosing if some of these fixes help solve some of the compilation issues.

travipross:
@mmatthe I've opened an additional PR in your fork for a simple addition to this PR regarding `boolean` types. 

@mvollrath Is the only outstanding issue holding up the approval of this PR related to the addition of the shortcut decoders in `rosbridge_library/src/rosbridge_library/internal/object_decoders.py`?

If so, is there any way we can move that portion to a separate PR for further discussion? There are some other fixes in this branch that address some currently broken functionality in the base branch, and it'd be nice to see this merged.

mvollrath:
> @mvollrath Is the only outstanding issue holding up the approval of this PR related to the addition of the shortcut decoders in `rosbridge_library/src/rosbridge_library/internal/object_decoders.py`?

Yes.

travipross:
Is there any way we can just comment/remove the two offending lines relating to checking for / returning a shortcut decoder, and potentially discuss their reintroduction in a different PR if desired?

It'd be nice to see the majority of this moved into the `ros2` branch, as in its current state, many features are still broken and this contains some fixes. 

I don't want to clutter things and open a new PR containing most of these same changes with this PR still open, but it seems @mmatthe might not be actively monitoring this PR right now.

AviKenz:
@mvollrath its seems that the change request was already proceeded. Please check [object_decoders.py](https://github.com/RobotWebTools/rosbridge_suite/blob/831b11630faa29b20d2f4e8a57cd1e4293b4e231/rosbridge_library/src/rosbridge_library/internal/object_decoders.py).
let me know if its enough or not. i will proceed the changes on my [PR](https://github.com/RobotWebTools/rosbridge_suite/pull/531/files) :)

travipross:
@mvollrath I've removed the offending `shortcut_object_decoders` in #533 as well as made a number of other small, but missing adaptations. I suggest we close this PR as @mmatthe seems to have abandoned it for now.

travipross:
@mvollrath It looks as though this has sat abandoned by the original author for almost a year and can be closed thanks to #533

PR #494, opened 2020-04-28
Added support for None for websocket_external_port

### LINKED ISSUE #489: External Port should be disabled when not set.
## Expected Behavior
If `websocket_external_port` is not set, then rosbridge should pass `None` to the autobahn library. This should disable the external port check. (https://github.com/crossbario/autobahn-python/issues/1360#issuecomment-614411125)

## Actual Behavior
The default value of `websocket_external_port` is set to the `port` value. This doesn't disable `externalPort` check even if it's not set. 

Let me know if you have any suggestions about this. I am working on a fix for this. I can submit a PR to fix this.

pramodhkp:
Fixes #489 

Also refer: https://github.com/crossbario/autobahn-python/issues/1360#issuecomment-614411125


rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -100,7 +100,12 @@ def shutdown_hook():
     port = rospy.get_param('~port', 9090)
     address = rospy.get_param('~address', "0.0.0.0")
 
-    external_port = int(rospy.get_param('~websocket_external_port', port))
+    external_port = rospy.get_param('~websocket_external_port', None)
+    if external_port:
+        try:
+            external_port = int(external_port)
+        except:
```
mvollrath, line 107: Please specific `except ValueError` here.

pramodhkp: Added

Issue #425, opened 2019-08-08
rosbridge_server hanging randomly under load


benmitch:
## Expected Behavior

rosbridge_server does not hang.

## Actual Behavior

rosbridge_server hangs under some load (several connections, quite high rates (50Hz), but relatively little data throughput).

## Steps to Reproduce the Problem

  1. Run rosbridge_server v 0.11.2 on an RPi 3B+
  1. Connect severally to it from a web browser.
  1. Wait.

## Specifications

  - ROS Version (`echo $ROS_DISTRO`): kinetic
  - OS Version (`grep DISTRIB_CODENAME /etc/lsb-release`): Raspbian GNU/Linux 9 (stretch)
  - Rosbridge Version (`roscat rosbridge_server package.xml | grep '<version>'`): 0.11.2
  - Tornado Version (`python -c 'import tornado; print tornado.version'`): 4.4.3

We are having trouble with rosbridge_server (0.11.2) freezing when under a fair load (number of connections on a somewhat flaky network, rather than throughput). The freeze appears to be random, and occurs after some period, typically 10-30 seconds.

I investigated, and it looks like it's a race condition for _write_lock, similar to that in issue #403, but not solved by the change of #413 (which is present at 0.11.2).

After some testing, and reading the issue reports, I noticed that somebody had suggested that the behaviour of "yield" inside a "with lock" section wasn't certain. So I tried:

    def prewrite_message(self, message, binary):
        # Use a try block because the log decorator doesn't cooperate with @coroutine.
        try:
            with self._write_lock:
                future = self.write_message(message, binary)

            # When closing, self.write_message() return None even if it's an undocument output.
            # Consider it as WebSocketClosedError
            if future is None:
                raise WebSocketClosedError

            yield future

That is, I un-indented after the line "future =", so the lock is released once the call has been made, but before the yield call.

Immediately, the behaviour seems to be fixed, and it has now been running robustly for several minutes without freezing again. I cannot comment at this stage on whether there have been any side effects, or data corruption, but the upstream signals look fine so far.

mvollrath:
[Somebody else was having issues with Tornado 4.4.3](https://github.com/RobotWebTools/rosbridge_suite/issues/422), you might try 4.5.3.  Yielding without the lock disconnects websocket backpressure and can lead to OOM on the server if it can't keep up with outgoing messages.  If your application is not very demanding, this is fine.

benmitch:
Thanks for your help with this.

[...]

PR #496, opened 2020-05-12
Fix backpressure deadlock

mvollrath:
See #425


J-Rojas:
I've tested these changes and they have solved the deadlock.

jihoonl:
Hey @J-Rojas , I see you frequently in PRs and comments in RWT. I invited in maintainers group. So, you can have more access to RWT repos. Thanks for the support.

PR #500, opened 2020-05-14
Threadsafe capabilities

mvollrath:
#496 fixed deadlock issues by decoupling incoming message handling from the web socket server, but this opened the door to race conditions when interacting with capabilities.  These changes add thread safety at the top-level which should prevent state inconsistencies further down.


mvollrath:
This was reported to cause hanging, so it's on hold for now.

PR #502, opened 2020-05-19
Finish protocol in IncomingQueue thread

mvollrath:
Prevents errors during protocol shutdown from breaking other states.


mvollrath:
These changes are [reported to work](https://github.com/RobotWebTools/rosbridge_suite/issues/425#issuecomment-632357597).

PR #506, opened 2020-05-29
Fixed filter_globs for noetic

foreignrobot:
Service calls with non empty requests (e.g. /rosapi/topics_for_type) were crashing due to filter's return type in python 3.

### DIFF
--- a/rosapi/src/rosapi/glob_helper.py
+++ b/rosapi/src/rosapi/glob_helper.py
@@ -2,9 +2,7 @@
 
 import rospy
 import fnmatch
-import sys
 
-PYTHON2 = sys.version_info < (3, 0)
 topics_glob = []
 services_glob = []
 params_glob = []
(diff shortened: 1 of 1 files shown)


rosapi/src/rosapi/glob_helper.py
```diff
@@ -30,7 +32,10 @@ def get_param(param_name):
 def filter_globs(globs, full_list):
     # If the globs are empty (weren't defined in the params), return the full list
     if globs is not None and len(globs) > 0:
-        return filter(lambda x: any_match(x, globs), full_list)
+        if PYTHON2:
+            return filter(lambda x: any_match(x, globs), full_list)
+        else:
+            return [i for i in filter(lambda x: any_match(x, globs), full_list)]
```
mvollrath: Instead of having all this logic specific to Python version, the Python 3 path should work fine in Python 2 and should be the only path if that is the case.

mvollrath (changes requested):

foreignrobot (commented):
Indeed the Python 3 path works fine in Python 2. Thus, I've removed the extra logic.

mvollrath (approved):
Thank you :+1:

hikinggrass:
@mvollrath is a noetic release with this fix planned in the near future (maybe for the next sync) or are there any other bugs you would like to get fixed before the next version bump?

mvollrath:
I don't see any reason not to tag a release, will do it soon after the new CI stuff is settled.

PR #507, opened 2020-06-03
Fix for type error caused by appending byte arrays to empty string

connormbrooks:
Fixes a TypeError caused when clients send incoming message as byte arrays. This error is introduced in ROS 2 due to Python 3's introduction of byte arrays distinct from strings.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/protocol.py
+++ b/rosbridge_library/src/rosbridge_library/protocol.py
@@ -118,7 +118,10 @@ class Protocol:
         message_string -- the wire-level message sent by the client
 
         """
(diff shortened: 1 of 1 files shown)


mvollrath:
It looks like if `self.buffer` is a non-zero-length `str` and a `bytearray` comes in it will raise a `TypeError` then continue to raise more `TypeError` for each `bytearray` until another `str` comes in to clear `self.buffer`.

connormbrooks:
Yes - this still retains the error if byte arrays and strings are intermixed, although it fixes the issue if a client is only sending one or the other.

mvollrath (approved):
Thank you :+1:

PR #517, opened 2020-08-08
is not to !=

flynneva:


rosapi/src/rosapi/params.py
```diff
@@ -221,7 +221,7 @@ def _get_param_names(node_name):
     rclpy.spin_until_future_complete(_node, future)
     response = future.result()
 
-    if response is not None:
+    if (response != None):
```
mvollrath: What is the purpose of this change?  Asking because using `is` to compare to `None` is correct.

mvollrath (changes requested):
Please remove extraneous parenthesis from these conditions.

rosapi/src/rosapi/params.py
flynneva: when I compile it on ubuntu 20.04 with foxy i get a stderror while compiling about this line.

```
[Processing: rosapi]                                    
--- stderr: rosapi                                       
/home/flynn/code/ros/foxy/install/rosapi/lib/python3.8/site-packages/rosapi/params.py:122: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if default is not "":
---
Finished <<< rosapi [49.6s]
```

flynneva: ill remove the parenthesis though

flynneva: that error is only for line 122, so im not sure if the second "is not" needs to be changed

mvollrath: Comparing to empty string value isn't the same as comparing to `None` which is a singleton.

flynneva: just trying to help out here...

mvollrath (approved):

mvollrath:
Thanks for the fix, should I be concerned about CI failures?

flynneva:
> Thanks for the fix, should I be concerned about CI failures?

I mean....it looks like it doesnt even get to building and testing your package for the ones that failed because it fails before for other packages (rosidl, i think). they might not be supported for those distros/OS?

Also though the ros ci action isnt that robust yet so maybe its an issue with the action itself rather than an actual build issue? I think i tried this package on eloquent and dashing before on linux and it worked ok (small setup issues, but solvable).

flynneva:
also i just realized i think the build errors might be what the other ROS2 Eloquent PR is for, right?

PR #518, opened 2020-08-08
add ROS2 github actions

flynneva:
- run only on ros2/* branches
- run for all ROS2 distros and platforms


flynneva:
example on my repo here: https://github.com/flynneva/rosbridge_suite/pull/2

mvollrath (approved):

PR #524, opened 2020-08-19
remove mac/windows from CI

flynneva:
mac & windows [not supported by the gh action](https://github.com/ros-tooling/action-ros-ci/issues/305) yet. will add it back in later once supported

bumped action version and added `vcs-repo-file-url` argument to ensure use of installed ROS2 binaries from previous step


mvollrath (approved):

mvollrath:
Thank you!

PR #525, opened 2020-08-19
add ros 1 github actions

flynneva:
- runs on PR to devel and master branches only
- runs on PR to ros1 and ros1/* branches only (no branches like this exist yet, but could in the future)

### DIFF
--- a/.github/workflows/ros_ci.yml
+++ b/.github/workflows/ros_ci.yml
@@ -1,0 +1,51 @@
+name: ROS 1 CI
+
+on:
+  pull_request:
+    branches:
+      - 'develop'
+      - 'master'
+      - 'ros1'
+      - 'ros1/*'
+jobs:
+  test_environment-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ros_ci.yml
```diff
@@ -0,0 +1,51 @@
+name: ROS 1 CI
+
+on:
+  pull_request:
+    branches:
+      - 'develop'
+      - 'master'
```
mvollrath: master branch isn't maintained anymore and should be omitted.

mvollrath (changes requested):

.github/workflows/ros_ci.yml
flynneva:
```suggestion
```

flynneva:
are there any ROS1 specific branches you'd like to run this on? we can add them to the triggers now

mvollrath:
develop is good for now.  Whenever it's time to reconcile porting across the branches we can name it proper.

mvollrath (approved):

PR #532, opened 2020-09-19
Fixes for Eloquent

travipross:
This PR is an extension of the currently open #493 that adds a number of similar fixes for type changes in ROS2.

This PR omits the inclusion of shortcut object decoders as noted in past reviews of #493.


travipross:
Apologies, there were some mistakes in this which will take a bit more tweaking

PR #537, opened 2020-10-13
rosbridge_udp now encodes msg before sending

kolibakter:
This is required to work with python3 socket library

Similar fix to #534 

Tested with ROS noetic (python3) and melodic (python2), seemed to be functional in both of them.


mvollrath (approved):

mvollrath:
Thank you!

PR #514, opened 2020-07-08
Possible Fix for #513

dignakov:
Possible fix for #513 

I ran into this today when trying to get RosSharp to work.

Everything works fine with roslibjs, but the same error as described in #513 came up when trying to use it with RosSharp (https://github.com/siemens/ros-sharp).

This seems to fix it for me...

PR #539, opened 2020-11-12
Fix unicode error for python2

vytasrgl:
Fixes the regression from #514 with unicode messages with python2


mvollrath:
Thank you for this patch and sorry for not giving this attention, I've merged #541 which should fix the error without using exceptions for flow control.

Issue #527, opened 2020-08-20
"Error processing request: field names must be a list or tuple type" when calling /rosapi/get_param_names in Python3


stevegolton:
/rosapi/get_param_names throws an exception when rosapi is running under Python 3.

Looks like the difference between the return type of `filter()` in python 2 vs python 3. In python 2 the list is evaluated immediately, but in python 3 filter returns a generator object which needs to be evaluated using the list() method or similar before the list is produced.

## Expected Behavior
Calling `/rosapi/get_param_names` should return the list of param names.

## Actual Behavior
`/rosapi/get_param_names` throws an exception. 

## Steps to Reproduce the Problem

  1. Run rosapi with python3.
  2. Call `/rosapi/get_param_names`
  3. See the following exception in the rosapi output:
```
[ERROR] [1597914827.181108]: Error processing request: field names must be a list or tuple type
[...]

PR #540, opened 2020-11-21
Evaluate filter() generator before returning in get_param_names()

stevegolton:
See issue #527.

This patch was needed when running rosapi under ROS Noetic which uses python3 due to the difference in in the return type of the `filter()` function between python2 and python3. In python2, a list is returned directly, but in python 3 a generator is returned which needs to be evaluated into a list to match the expected return type of the `get_param_names()` function in `params.py`.

Also tested using the ros-melodic docker container which uses python2 and it seems to work fine.

Please let me know if this is the wrong place to submit pull requests to, or if the python3/noetic migration work is being done elsewhere?

### DIFF
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -110,7 +110,7 @@ def get_param_names(params_glob):
     with param_server_lock:
         if params_glob:
             # If there is a parameter glob, filter by it.
-            return filter(lambda x: any(fnmatch.fnmatch(str(x), glob) for glob in params_glob), rospy.get_param_names())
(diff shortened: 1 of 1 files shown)


mvollrath:
This is a good place for it, we've been keeping Python 2/3 compatibility in noetic.

It looks like the tests are choking on some ROS test infrastructure, maybe something in [here](https://github.com/ros-tooling/setup-ros), but I don't see an issue tracked yet.

mvollrath:
Tests are passing for me.

stevegolton:
Thanks @mvollrath! I was struggling a little with those tests.

PR #541, opened 2020-11-24
Use Python 3 str in incoming

mvollrath:
Fixes Python 2 unicode issues.


mvollrath:
Supersedes #539 to fix regression in #514

moriarty:
This adds a dependency on `python-future` which did not exist before and breaks at runtime when it is not installed

PR #548, opened 2021-01-01
fixed SyntaxWarning: "is not" with a literal

danidask:
It compiles in ros2 foxy and ubuntu 20.04


mvollrath (approved):

jtbandes:
Fixed in #574. Thanks for the submission!

Issue #553, opened 2021-01-28
Duplicate servers returned from /rosapi/action_servers


alexytsu:
## Expected Behavior

Calling the `rosapi/action_servers` service should not return duplicate action servers.

## Actual Behavior

Duplicate action servers can be returned after querying from roslibjs.

## Steps to Reproduce the Problem

1. `rosrun turtle_actionlib shape_server` or any other Action server
2. `roslaunch rosbridge_server 
2. `rosservice call /rosapi/action_servers "{}"`:  Returns: action_servers: [/turtle_shape] (Correct behaviour)
4. Now from the the `roslibjs` client library call `getActionServers` which will return ["/turtle_shape", "/turtle_shape"] (Incorrect behaviour in roslibjs client library)
5. `rosservice call /rosapi/action_servers "{}"`: Now alos returns  action_servers: [/turtle_shape, /turtle_shape] (Incorrect behaviour over the raw service call as well now)

## Specifications

  - ROS Version (`echo $ROS_DISTRO`): melodic
  - OS Version (`grep DISTRIB_CODENAME /etc/lsb-release`): DISTRIB_CODENAME=bionic
  - Rosbridge Version (`roscat rosbridge_server package.xml | grep '<version>'`): <version>0.11.13</version>
  - Twisted Version (`python -c 'import twisted; print twisted.version'`): [Twisted, version 17.9.0]

alexytsu:
Ok the actual reproduction for this is simpler than I first thought. What actually causes the bug to occur is if there are topics published that are not long enough to be caught by the `if(len(split) >= 3)` check. So in my case I had the topic `waypoints` published by my dashboard (which was using roslibjs).

In these cases, if an action server was just found, the `possibility` array is not cleared and the next iteration of the loop will pass the ` if all(p == 1 for p in possibility):` check and add the previous possible_action_server to the array again.

I've proposed a simple fix https://github.com/RobotWebTools/rosbridge_suite/pull/554#issue-562915563

PR #554, opened 2021-01-28
Fix duplicate action servers showing up

alexytsu:
This is one potential fix for the issue described in https://github.com/RobotWebTools/rosbridge_suite/issues/553


alexytsu:
Unsure why this actually works as the list of topics passed in to `filter_action_servers` seems to remain the same in the buggy case.

alexytsu:
@mvollrath is it possible to get this PR either reviewed or closed as a won't fix so that we can accommodate on client side code appropriately?

PR #556, opened 2021-02-03
Fix for python2 regression from merge of #514

J-Rojas:
Fix for bson_only_mode, avoid concatentation of ascii buffered bytes.


mvollrath:
@J-Rojas if you would like to continue support of BSON only mode, basic end-to-end regression tests (similar to the rosbridge_server "smoke" tests) are called for to prevent this kind of regression.

J-Rojas:
Okay I'll look into it.

PR #561, opened 2021-02-18
Fix invalid conversion bug in time and duration

buster84:
There is a invalid conversion bug in time and duration. While I use roslibjs, the `sec` and `nsecs` keys become `'115,101,99,115'` and `'110,115,101,99,115'`. This pr fixes the problem.


mvollrath (approved):
Thank you!

buster84:
Hi, thank you for the fixing flake8. Can you merge this and release a new version?

PR #562, opened 2021-03-03
Update subscribers.py

hangst:
`.values()` should return an iterable. Wrapping with another iterable (list) leads to this error.

```
[ERROR] Exception calling subscribe callback: 'dict_values' object is not callable
```

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -204,7 +204,7 @@ class MultiSubscriber():
         subscriptors.
         """
         with self.lock:
-            self.callback(msg, [self.new_subscriptions.values()])
+            self.callback(msg, self.new_subscriptions.values())
             self.subscriptions.update(self.new_subscriptions)
             self.new_subscriptions = {}
             self.node_handle.destroy_subscription(self.new_subscriber)


jtbandes:
Hi, can you share some more information and reproduction steps for the bug that's being fixed here? I've added a smoke test in #585 and it would be great if you could update this PR to ensure this code is covered by the tests.

hangst:
With `rosbridge_websocket.py` (ros2 branch) running, this script should trigger `'dict_values' object is not callable` error. 

```html
<script>
const payload = JSON.stringify({
    "op": "subscribe",
    "topic": "/rosout",
})
new WebSocket("ws://localhost:9090").addEventListener("open", function () {
    this.send(payload) // 1st subscription request
    this.send(payload) // 2nd subscription request, triggers error
})
</script>
```

This error happens when a WebSocket client makes multiple subscription requests on the same topic.

hangst:
@jtbandes FYI this is the Dockerfile I used to build `rosbridge_suite` from source. 

```Dockerfile
FROM ros:foxy
WORKDIR /root
RUN git clone --depth 1 --branch ros2 https://github.com/RobotWebTools/rosbridge_suite
RUN apt-get update
RUN rosdep update
RUN rosdep install --default-yes --ignore-src --from-path .
RUN . /opt/ros/foxy/setup.sh && colcon build
CMD ["bash", "-c", ". install/setup.bash && ros2 run rosbridge_server rosbridge_websocket.py"]
```

`rosbridge_websocket.py` is executed with the following command. 

```bash
docker build . --tag rosbridge_suite
docker run -it --net=host rosbridge_suite
```

jtbandes:
Thanks for the contribution. I am going to close this PR and work from https://github.com/RobotWebTools/rosbridge_suite/pull/671 instead, which covers the same problem.

Issue #536, opened 2020-10-08
Foxy Release?


JWhitleyWork:
## Expected Behavior
This stack is availaable on the ROS Build Farm for supported versions of ROS2.

## Actual Behavior
A `dashing` release exists but not a `foxy` one.

## Dependencies that Need Releasing
- https://github.com/GT-RAIL/rosauth/

JWhitleyWork:
I have verified that the stack builds with the source version of `rosauth` on the `ros2` branch.

JWhitleyWork:
`rosauth` now has [a release](https://github.com/ros/rosdistro/pull/27767) [pending sync](http://repo.ros2.org/status_page/ros_foxy_default.html?q=SYNC). Can the maintainer of this repo (@mvollrath ?) please consider a ROS Foxy release from the `ros2` branch?

JWhitleyWork:
Ping to @mvollrath for Foxy release. `rosauth` is now available in the repositories.

mvollrath:
Would like to see a new distro appear in CI tests before releasing the package.

JWhitleyWork:
@mvollrath The `ros2` branch already has Foxy jobs, but only on PRs. See https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/.github/workflows/ros_ci.yml.

sageig:
Just wanted to note that the Foxy job will build successfully if the setup-ros and action-ros-ci versions are updated from 0.0.25 and 0.0.19, respectively, to 0.1.0. I tested it in [this PR](https://github.com/RobotWebTools/rosbridge_suite/pull/557). However, the dashing and eloquent jobs are failing in the same manner as the kinetic and melodic jobs.

sageig:
@mvollrath Does the above suffice, or are there other things you're looking for that could help us move this forward?

jtbandes:
Hi folks, what was the status of this? What will it take in order to get rosbridge-suite released for Galactic/Humble?

amacneil:
This was discussed in the ros web tools working group today, and will happen soon.

jtbandes:
- rolling: https://github.com/ros/rosdistro/pull/30483
- galactic: https://github.com/ros/rosdistro/pull/30479
- foxy: https://github.com/ros/rosdistro/pull/30480

JWhitleyWork:
Thank you!

amacneil:
[...]

PR #574, opened 2021-07-16
Add cbor-raw compression support (ros2)

jtbandes:
This PR is an adaptation of #452 for the `ros2` branch.

After the initial cherry-pick, a new implementation of the `/rosapi/get_topics_and_raw_types` service was required for ROS 2, because the generated Python msg classes don't contain the original/combined message definition (see related rosbag2 discussion at https://github.com/ros2/rosbag2/issues/782). To accomplish this, I added a `stringify_field_types()` which reconstitutes a plausible .msg definition (although without constants or default values) based on the `_fields_and_field_types`, in the `====`-separated format seen in ROS 1 bags, which can be parsed by ROS 2 .msg parsers such as [@foxglove/rosmsg](https://github.com/foxglove/rosmsg).

I hope that `stringify_field_types` could be removed in the future if message generation is modified to output the original message definition text (https://github.com/ros2/ros2/issues/1159) — I see it as a stopgap solution to get basic compatibility working.

I also modified `get_publications_and_types` to remove the extra `/msg/` from the ROS 2 IDL names, so `std_msgs/msg/String` becomes `std_msgs/String` when the list of topics and datatypes are retrieved. This is more in line with what clients migrating from ROS 1 expect.

I had to make a few launch file / parameter changes to get the `rosbridge_websocket_launch.xml` to work. The test included with #452 is not enabled, because there seems to be no `rostest` replacement for ROS 2 and I wasn't sure how best to port the test. I noticed that some other tests are also disabled in this branch.

@mvollrath or @dirk-thomas, would you be able to help review this? I'd also be interested in pursuing including this package into Foxy/Galactic/Humble. It looks like it was released in Dashing but has been dropped recently (#536) and I'm not clear if that was intentional, or if it might just take a little shepherding to get it back into a release again.

(I'm also curious if you think there is any future where the ROS 1 and 2 versions of this package might be able to coexist, rather than living as two separate branches forever. It seems like there have already been several changes on develop that aren't in the ros2 branch...)


dirk-thomas:
Sorry, I am not involved in the project anymore.

jtbandes:
If you happen to know who is currently involved, I’d appreciate a pointer! Thanks

mvollrath:
> (I'm also curious if you think there is any future where the ROS 1 and 2 versions of this package might be able to coexist, rather than living as two separate branches forever. It seems like there have already been several changes on develop that aren't in the ros2 branch...)

Major changes were made to ros1 develop after ros2 was branched, including fixes for backpressure and deadlock conditions.  To summarize, in current ros2 websocket server outgoing messages are infinitely queued by Tornado *after* the implemented rosbridge_library queue/throttle features.  This means a client can cause OOM on the server by subscribing to topics and not handling the messages fast enough.

The fix:
* Outgoing message writes from rosbridge_library are blocked by the WebSocket server until its buffer is below some reasonable level.  This newly-connected backpressure is managed by rosbridge_library queue/throttle features.
* Under some conditions incoming messages can immediately produce outgoing messages, these conditions must be thread-decoupled to prevent deadlocking the WebSocket server while buffer is blocking.

jtbandes:
Do you see an easy path to backport (forward-port?) these changes to the ros2 branch?

Was there any particular reason the ros2 branch wasn't published in a release since Dashing? I'd love to get this feature in and get it published in some or all of Foxy/Galactic/Humble if possible.

mvollrath:
> Do you see an easy path to backport (forward-port?) these changes to the ros2 branch?

Potentially just re-apply all the PRs, as long as there's E2E tests on the ros2 branch (like there are in ros1 develop) this should be low risk.

> Was there any particular reason the ros2 branch wasn't published in a release since Dashing? I'd love to get this feature in and get it published in some or all of Foxy/Galactic/Humble if possible.

Not that I know of, just need somebody to push the button.  I don't have any reason to take ownership of this atm.

jtbandes:
Yep, re-applying the PRs makes sense. Unfortunately it looks like the e2e tests [were disabled](https://github.com/RobotWebTools/rosbridge_suite/commit/7b4a8d9f2273afaa1156b9683811bdcc1d125f94#diff-9880ecef4f0b51f9f2ddac1d64e89f9e9efbd11e74bb3926eaaeeaa332838143R43) in one of the initial commits porting to ROS 2. If someone (@dirk-thomas?) could point me at a suitable replacement for rostest, I could work on re-enabling the tests.

Part of my question was also: do you think it's possible to make ros1 and ros2 releases from the same branch of the same repo? That way we wouldn't continue having feature disparities between the two branches. Alternatively, would it make sense to make ros2 the default branch, and treat the ros1 develop branch as bugfix-only?

mvollrath:
> Yep, re-applying the PRs makes sense. Unfortunately it looks like the e2e tests [were disabled](https://github.com/RobotWebTools/rosbridge_suite/commit/7b4a8d9f2273afaa1156b9683811bdcc1d125f94#diff-9880ecef4f0b51f9f2ddac1d64e89f9e9efbd11e74bb3926eaaeeaa332838143R43) in one of the initial commits porting to ROS 2. If someone (@dirk-thomas?) could point me at a suitable replacement for rostest, I could work on re-enabling the tests.

I think [launch_testing](https://github.com/ros2/launch/tree/master/launch_testing) is current practice.

> Part of my question was also: do you think it's possible to make ros1 and ros2 releases from the same branch of the same repo? That way we wouldn't continue having feature disparities between the two branches. Alternatively, would it make sense to make ros2 the default branch, and treat the ros1 develop branch as bugfix-only?

Don't think it's practical to use the same branch.  Would make sense for ros2 to be default, hopefully everybody understands ROS1 branch is maintenance only since we're now in the final LTS distro.

jtbandes:
(or maybe rename develop to ros1, and ros2 to develop, or whatever is needed to bring this repo in line with standard branch naming for ROS projects?)

jihoonl:
I have been in charge of the release process for ros1. 

I don't think anyone is in charge of ros2 at the moment. If anyone with proper background steps up for ros2 release responsibility, I am happy to help you.

jtbandes:
@jihoonl Thanks for the reply. Foxglove would be happy to take responsibility for the ros2 release, but will definitely need your help 🙂  Happy to coordinate here or in [our Slack](http://foxglove.dev/join-slack) or ROS Discord. I and/or some others from Foxglove will also be at the web tools working group meeting this week.

jihoonl:
@jtbandes I don't think I am joining the next web tools group meeting. Would you send me an email with one or two github ids on it? I will invite them as maintainers for ros2 branch.

> Don't think it's practical to use the same branch. Would make sense for ros2 to be default, hopefully everybody understands ROS1 branch is maintenance only since we're now in the final LTS distro.
> (or maybe rename develop to ros1, and ros2 to develop, or whatever is needed to bring this repo in line with standard branch naming for ROS projects?)

I agree that we should set ros2 branch as default at some point. But, I am unsure whether it is the right time to do it.

esthersweon:
@jihoonl - Hi there, I'm another Foxglove team member. Just sent you an email with the info, as @jtbandes is out this week. Please let me know how I can help get this over the finish line.

Issue #551, opened 2021-01-22
QoS compatibility failing silently (ROS2)


DomenicP:
**Edit**: I originally thought the problem I encountered was related to custom message definitions, but it turns out that the problem was actually regarding ROS2 QoS settings. As such I have updated the issue description.

## Expected Behavior

`rosbridge_server` should allow clients to subscribe to a topic regardless of the QoS settings for that topic.

## Actual Behavior

`rosbridge_server` silently fails to subscribe to topics with publishers that use **Best Effort** reliability instead of **Reliable.** The `rosbridge_server` logs indicate that the client connected and that the subscription was created, but no data is ever received in the subscriber callback.

```
[INFO] [1611352924.485146047] [rosbridge_websocket]: Client connected. 1 clients total.
[INFO] [1611352926.636802670] [rosbridge_websocket]: [Client 1] Subscribed to /my_topic
```

According to the [QoS documentation](https://index.ros.org/doc/ros2/Concepts/About-Quality-of-Service-Settings/#qos-compatibilities), "a connection between a publisher and a subscription is only made if the pair has compatible QoS profiles." For the reliability QoS policy, all combinations are compatible except for when the publisher uses **Best Effort** and the subscriber uses **Reliable**. Based on my recent experience, it would seem that a failure to connect based on QoS compatibility either fails silently or requires more verbose logging than I currently have configured. 

The rosbridge subscriptions are initially created here:

https://github.com/RobotWebTools/rosbridge_suite/blob/1a2c386455067b00a9e67fb70cc711add38c3582/rosbridge_library/src/rosbridge_library/internal/subscribers.py#L108

It looks like `rosbridge_server` uses the default reliability QoS policy (which in my case ends up being **Reliable**) and only specifies the depth. Would it be better to have the rosbridge subscribers use a **Best Effort** QoS policy? I'm not super familiar with how things work behind the scenes, but from what I read in the documentation it sounds like a subscriber that requests **Best Effort** from a publisher that provides **Reliable** will still end up with **Reliable** QoS. The rosbridge protocol makes no mention of QoS since it's a new concept in ROS2, so the client doesn't currently have a way to request a specific QoS profile.

## Steps to Reproduce the Problem

  1. Run a ROS2 node that publishes data to a topic using **Best Effort** reliability.
  2. Run `rosbridge_server` and attempt to subscribe to that topic using a rosbridge client library (roslibpy, roslibjs). The subscriber should not receive any data.
[...]

PR #582, opened 2021-08-09
ROS2 - Add qos_durability (volatile or transient_local) option to subcriptions

MrBlenny:
Inspired by https://github.com/RobotWebTools/rosbridge_suite/pull/557 This PR would break when a topic was being published as `volatile` but being subscribed as `transient_local` - as described [here](https://docs.ros.org/en/foxy/Concepts/About-Quality-of-Service-Settings.html#qos-compatibilities)

To resolve this, I have made this configurable using a `qos_durability` parameter.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/subscribe.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/subscribe.py
@@ -92,7 +92,8 @@ class Subscription():
         self.clients.clear()
 
     def subscribe(self, sid=None, msg_type=None, throttle_rate=0,
-                  queue_length=0, fragment_size=None, compression="none"):
+                  queue_length=0, qos_durability="volatile", fragment_size=None, 
+                  compression="none"):
         """ Add another client's subscription request
 
         If there are multiple calls to subscribe, the values actually used for
@@ -106,6 +107,7 @@ class Subscription():
         being sent.  If multiple subscriptions, the lower of these is used
         queue_length    -- the number of messages that can be buffered.  If
         multiple subscriptions, the lower of these is used
+        qos_durability  -- the qos durability, 'volatile' or 'transient_local'
         fragment_size   -- None if no fragmentation, or the maximum length of
         allowed outgoing messages
         compression     -- "none" if no compression, or some other value if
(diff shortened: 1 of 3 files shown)


DomenicP:
I ran into a similar QoS issue earlier this year (#551) except with the reliability setting instead of durability. I like your approach of making the durability a parameter; perhaps we could do something similar for the reliability as well? I'd be happy to contribute the code changes if it falls under the scope of this PR or I can open a new PR after this one is approved.

jtbandes:
This implementation suffers from a bug similar to the one I unwittingly introduced in #574 and subsequently filed as https://github.com/RobotWebTools/rosbridge_suite/issues/610 — since multiple client subscriptions are merged into a single MultiSubscriber instance, the qos setting from whichever client subscribes first will take precedence rather than correctly creating two subscriptions with different qos settings.

I am inclined to think we should just remove the MultiSubscriber since it adds complexity. (There is another bug around MultiSubscribers, with an open PR to fix it at https://github.com/RobotWebTools/rosbridge_suite/pull/562 – I haven't had time to test and evaluate the fix yet.)

kenji-miyake:
How about getting qos_profile before subscribing? :thinking: https://github.com/tier4/rosbridge_suite/pull/6
Perhaps it might require some error handlings(e.g. no topic after rosbridge launch), but it worked in my simple local environment.

It seems `ros2 topic echo` does the same thing.
https://github.com/ros2/ros2cli/blob/d974fdefb0e8f0af095ac6e211e5b17b0115974a/ros2topic/ros2topic/verb/echo.py#L140

kenji-miyake:
@DomenicP As for `qos-reliability`, I thought the re-connection feature can solve it. https://github.com/tier4/rosbridge_suite/pull/7
How do you feel about it?

DomenicP:
@kenji-miyake I was unfamiliar with those changes, thanks for bringing them to my attention. I'll give your PR a try and see if that fixes the issue I ran into earlier this year.

defunctzombie:
I think we should go with either #654 - or some variant of #689 which detects if one should use volatile or transient_local for the user.

I found it frustrating and non-obvious when subscribing to `/tf_static` and not receiving the latched messages. The default case should be subscribe and get the published message. There might be some interesting cases to expose overriding the profile but the default behavior should be more intuitive.

PR #593, opened 2021-08-18
Fix some but not all Python undefined names

cclauss:
Use the test `flake8 --select=F82` to find and fix _undefined names_ in Python code like [we used to do in Travis CI](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/.travis.yml#L35).  Several of these changes are about removing Python 2 data types that were dropped in Python 3 (basestring, long, unicode).  Other fixes were about missing imports or local variable naming.  This PR fixes 24 of 26 _undefined name_ but was unable to resolve the issues in the two excluded files.   #604

## After:
$ `flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics`
```
./rosapi/src/rosapi/proxy.py:240:11: F821 undefined name 'get_service_uri'
    uri = get_service_uri(service)
          ^
./rosbridge_server/scripts/rosbridge_websocket.py:304:5: F821 undefined name 'node'
    node.destroy_node()
    ^
2     F821 undefined name 'get_service_uri'
2
```

## Before:
$ `flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics`
```
./rosbridge_library/test/internal/test_services.py:19:26: F821 undefined name 'unicode'
    string_types = (str, unicode)
                         ^
./rosbridge_library/test/internal/test_services.py:30:53: F821 undefined name 'unicode'
    elif sys.version_info < (3,0) and isinstance(d, unicode):
                                                    ^
./rosbridge_library/test/internal/test_services.py:31:16: F821 undefined name 'unicode'
        return unicode(random.random())
               ^
./rosbridge_library/test/internal/test_message_conversion.py:21:26: F821 undefined name 'unicode'
    string_types = (str, unicode)
                         ^
./rosbridge_library/test/internal/test_cbor_conversion.py:39:55: F821 undefined name 'unicode'
            self.assertEqual(type(extracted['data']), unicode)
                                                      ^
./rosbridge_library/test/internal/test_cbor_conversion.py:152:45: F821 undefined name 'unicode'
                self.assertEqual(type(key), unicode)
                                            ^
./rosbridge_library/src/rosbridge_library/capabilities/subscribe.py:322:21: F821 undefined name 'unicode'
            topic = unicode(topic)
                    ^
./rosbridge_library/src/rosbridge_library/util/__init__.py:16:26: F821 undefined name 'unicode'
    string_types = (str, unicode)
                         ^
[...]

PR #594, opened 2021-08-18
GitHub Action to lint Python code

cclauss:
Because Travis CI seems to be on (permanent?) vacation...
https://travis-ci.org/github/RobotWebTools/rosbridge_suite

Related to #593 and #600

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.


amacneil:
These are a lot of steps if someone wants to run lint locally. Can we move it to a configuration file to simplify CI and local linting?

cclauss:
We could move much of this to a pre-commit that can be run locally and in CI.

cclauss:
The four lines from 10 thru 13 should be equivalent on a local machine.

.github/workflows/ci.yml
```diff
@@ -3,6 +3,19 @@ name: CI
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip
+      - run: pip install bandit codespell flake8
+      - run: bandit --recursive --skip B101,B110,B311 .
+      # See setup.cfg for the default settings for codespell and flake8
+      - run: codespell
+      - run: flake8
+      # Undefined names need to be fixed (#604) and the --exit-zero removed.
+      - run: flake8 . --count --exit-zero --select=F82 --show-source --statistics
```
amacneil: no need for `show-source` here since it's in our `setup.cfg` below

amacneil: also I'm pretty sure `.` is unnecessary? it seems to run on `.` by default

.github/workflows/ci.yml
```diff
@@ -3,6 +3,19 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip
```
amacneil: I don't think this line is necessary? `actions/setup-python` already gives us the latest version (see output from this on your branch)

.github/workflows/ci.yml
```diff
@@ -3,6 +3,19 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
```
amacneil: Can you change this to `ubuntu-20.04` so we don't get a surprise upgrade

.github/workflows/ci.yml
```diff
@@ -3,6 +3,19 @@ name: CI
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
```
amacneil, line 10: should we specify a python version here? I assume that could affect the lint results.

I don't know what the minimum python version supported by ros 2 is, but I assume `3.8` is a safe choice since that's what comes with ubuntu 20.04 (used by foxy and galatic)

setup.cfg
```diff
@@ -0,0 +1,9 @@
+[codespell]
+ignore-words-list = miror
+
+[flake8]
+count = True
+max-complexity = 36
+max-line-length = 228
+select = E5,E9,F63,F7
```
amacneil: - I think you need to add `C9` here for the cyclomatic complexity check to work? (at least I did on my machine - try setting `max-complexity` to something lower to see what I mean)
- what do you think about including all `F` checks (and excluding the failing ones for now) like I did in https://github.com/RobotWebTools/rosbridge_suite/pull/605/files#diff-6951dbb399883798a226c1fb496fdb4183b1ab48865e75eddecf6ceb6cf46442R2-R4 ?

.github/workflows/ci.yml
cclauss: Not always...  Watch it over time.

.github/workflows/ci.yml
cclauss: Don't we want to know (before our users do) that Python 3.10 breaks us?

setup.cfg
cclauss: I think that to F issues should be fixed because PEP8 is clear about wildcard imports being a bad idea.

.github/workflows/ci.yml
amacneil: Good question. I'm not sure what is common, but I would think that most people are using the version of python that comes with ubuntu (since ROS requires a specific ubuntu distribution)?

If we really care about that then it would be best to test both the minimum and maximum versions. But for lint it's probably safer to run on the minimum supported version?

Open to ideas.

amacneil (approved):
Thanks!

setup.cfg
amacneil: sorry I don't understand - can you rephrase?

PR #598, opened 2021-08-18
Remove Dashing and Eloquent from CI; add Rolling

jtbandes:
Dashing and Eloquent are EOL and [not accepting new releases](https://discourse.ros.org/t/error-running-bloom-release-for-eloquent-track-of-rosbridge-suite/21834/2?u=jtbandes). Removing them from CI allows us to start using newer APIs (e.g. in https://github.com/RobotWebTools/rosbridge_suite/pull/597).

PR #599, opened 2021-08-18
Clean up Github CI action

amacneil:
Dashing and Eloquent are EOL and [not accepting new releases](https://discourse.ros.org/t/error-running-bloom-release-for-eloquent-track-of-rosbridge-suite/21834/2?u=jtbandes). Removing them from CI allows us to start using newer APIs (e.g. in https://github.com/RobotWebTools/rosbridge_suite/pull/597).

Closes #598


.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: CI
+
+on:
+  push:
+    branches: ["ros2"]
+  pull_request:
+    branches: ["*"]
```
jtbandes, line 7: Seems like this would run for PRs on the ros1 branch as well, right? That's why it was previously set to ros2

jtbandes (approved):
Looks good although I'm not sure about the `"*"` branch selector for the pull_request trigger.

.github/workflows/ci.yml
amacneil: I'm pretty certain that the action yml from the PR branch is what github runs, not the yml from the default or target branch. So we just need to put the appropriate configuration in ros1 and ros2 branches and not worry about this branches wildcard.

PR #605, opened 2021-08-19
Add flake8 configuration file and lint to CI

amacneil:
Builds on top of https://github.com/RobotWebTools/rosbridge_suite/pull/594, plus:

- Moves flake8 configuration to a file so people can just run `flake8` locally
- Add to existing Github Action `ci.yml` rather than creating a new one

cc @cclauss any feedback?


cclauss:
See #606 which removes the `+` sign at the end of https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/rosbridge_library/src/rosbridge_library/internal/publishers.py#L226

The same problem repeats 4 and 5 lines down.

cclauss:
Perhaps instead of a `.flake8` file, go for a `setup.cfg` file that has sections for ~bandit,~ codespell, and flake8.

.github/workflows/ci.yml
```diff
@@ -34,3 +34,19 @@ jobs:
       - uses: ros-tooling/action-ros-ci@v0.2
         with:
           target-ros2-distro: ${{ matrix.ros_distribution }}
+
+  lint:
```
cclauss, line 38: Maybe run lint first as a fast-fail given that it only take ~20 sec.

cclauss:
`setup.cfg` ideas... https://github.com/RobotWebTools/rosbridge_suite/pull/594/files

.github/workflows/ci.yml
amacneil: They all run in parallel so people will get lint feedback quickly either way. I tend to prefer not failing fast - Github actions are free for open source so the extra resources aren't costing us anything, and it can save a round trip if you know that you have to fix both a lint issue and a build/test issue (rather than fixing a lint issue only to find afterwards that you also have a failing test).

amacneil:
Thanks - let's iterate on feedback in https://github.com/RobotWebTools/rosbridge_suite/pull/594

PR #609, opened 2021-08-19
Fix wildcard imports and unused variables

cclauss:
Fix some flake8 F401,F403, and F841 errors.

### DIFF
--- a/rosapi/src/rosapi/proxy.py
+++ b/rosapi/src/rosapi/proxy.py
@@ -151,7 +151,7 @@ def get_topic_type(topic, topics_glob):
     topics, types = get_topics_and_types(topics_glob)
     try:
         return types[topics.index(topic)]
-    except ValueError as e:
+    except ValueError:
         # Return empty string if the topic is not present.
         return ""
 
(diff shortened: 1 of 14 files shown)


jtbandes (approved):

cclauss:
The wildcard imports that I could not figure out are at https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/rosbridge_library/test/internal/publishers/test_publisher_consistency_listener.py#L9-L11  They hide the _undefined name_ [`PublisherConsistencyListener`](https://github.com/RobotWebTools/rosbridge_suite/search?q=PublisherConsistencyListener) which is not imported or defined in this repo.  Is this a dead test?

jtbandes:
Looks like PublisherConsistencyListener was deleted in #419. The tests were disabled in https://github.com/RobotWebTools/rosbridge_suite/commit/7b4a8d9f2273afaa1156b9683811bdcc1d125f94.

From that PR description:

> In publishers module inside internal, a ROS1 SubscriberListener was used to prevent late-joining subscribers from missing messages. This is now achieved using transient local QoS with a lifespan.

So I suspect it is safe to delete the test. Perhaps @jubeira can confirm.

PR #613, opened 2021-08-19
Delete test_publisher_consistency_listener.py

cclauss:
This file no longer seems to be required as discussed at https://github.com/RobotWebTools/rosbridge_suite/pull/609#issuecomment-902152912

@jubeira Your review, please.


jtbandes:
The reference to this test should also be removed from `rosbridge_library/test/internal/test_internal.test`

jtbandes (approved):

PR #614, opened 2021-08-19
Python idioms and is for singletons

cclauss:
https://docs.python.org/3/library/2to3.html#2to3fixer-idioms plus equality vs. identity fixes (flake8 F71x issues).

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
@@ -58,7 +58,7 @@ class Defragment(Capability, threading.Thread):
         self.protocol = protocol
 
         # populate parameters
-        if self.protocol.parameters != None:
+        if self.protocol.parameters is not None:
             self.fragment_timeout = self.protocol.parameters["fragment_timeout"]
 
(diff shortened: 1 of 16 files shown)


jtbandes (approved):

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
```diff
@@ -106,9 +106,7 @@ def defragment(self, message):
         msg_data = message.get("data")
 
         # Abort if any message field is missing
-        if ((msg_opcode == None) or (msg_id == None) or
-            (msg_num == None) or (msg_total == None) or
-            (msg_data == None)):
+        if None in (msg_opcode, msg_id, msg_num, msg_total, msg_data):
```
jtbandes, line 109: Is this a common Python idiom? I haven't seen it before, but it does seem more concise than several `is None`.

jtbandes (approved):
Thanks!

rosbridge_library/src/rosbridge_library/capabilities/defragmentation.py
cclauss: It is a pragmatic solution rather than a common solution.

PR #621, opened 2021-08-20
Miscellaneous fixes: flake8 --select=E402,E731,E741,F821

cclauss:
Fix six miscellaneous flake8 issues that formatters like [psf/black](https://github.com/psf/black) will not auto-fix.
```
2     E402 module level import not at top of file
1     E731 do not assign a lambda expression, use a def
1     E741 ambiguous variable name 'l'
2     F821 undefined name 'get_service_uri'
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,6 @@ jobs:
       # See setup.cfg for the default settings for codespell and flake8
       - run: codespell
       - run: flake8
-      # Undefined names need to be fixed (#604) and the --exit-zero removed.
-      - run: flake8 --exit-zero --select=F82 --statistics
   test:
     strategy:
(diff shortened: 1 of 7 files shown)


amacneil:
Can you add all the fixed checks to `setup.cfg` in these PRs so we can prevent future regressions?

cclauss:
Let's just land these PRs and run psf/black and we will pass ALL flake8 tests.

cclauss:
Adding them is going to mean that each PR has conflicts with all others because they are all modifying the same line.

amacneil:
> Adding them is going to mean that each PR has conflicts with all others because they are all modifying the same line.

I see, yes makes sense

jtbandes (approved):

PR #622, opened 2021-08-20
Do not use bare 'except:': flake8 --select=E722

cclauss:
Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on. Prefer `except Exception:`. If you’re sure what you’re doing, be explicit and write `except BaseException:`.

https://realpython.com/the-most-diabolical-python-antipattern/
https://www.python.org/dev/peps/pep-0008/

### DIFF
--- a/rosbridge_library/test/capabilities/test_subscribe.py
+++ b/rosbridge_library/test/capabilities/test_subscribe.py
@@ -56,11 +56,8 @@ class TestSubscribe(unittest.TestCase):
             self.assertEqual(subscription.queue_length, min_queue_length)
             self.assertEqual(subscription.fragment_size, min_frag_size)
             self.assertEqual(subscription.compression, "png")
-        except:
+        finally:
             subscription.unregister()
-            raise
-
-        subscription.unregister()
 
     def test_missing_arguments(self):
         proto = Protocol("test_missing_arguments")
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -108,7 +108,7 @@ def _set_param(node_name, name, value, parameter_type=None):
     try:
         # call_get_parameters will fail if node does not exist.
         call_set_parameters(node=_node, node_name=node_name, parameters=[parameter])
-    except:
+    except Exception:
         pass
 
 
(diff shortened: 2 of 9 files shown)


jtbandes (approved):

rosbridge_library/test/capabilities/test_subscribe.py
```diff
@@ -56,7 +56,7 @@ def test_update_params(self):
             self.assertEqual(subscription.queue_length, min_queue_length)
             self.assertEqual(subscription.fragment_size, min_frag_size)
             self.assertEqual(subscription.compression, "png")
-        except:
+        except:  # noqa: E722  # Will unregister and raise
             subscription.unregister()
             raise
```
jtbandes: Should these use `finally:` instead of `except: ... raise`?

cclauss: I think not...

 % `python3`
```python
>>> try:
...     print("Hello")
... finally:
...     raise
...
Hello
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
RuntimeError: No active exception to reraise
```
We want the exception to propagate up the call stack.

jtbandes: sorry, I meant 
```py
try:
  ...
finally:
  subscription.unregister()
```

cclauss: ~I think we are good the way it is.  If failure unregister and raise else unregister (the line after) does what we want.~

cclauss: Oh...
```python
try:
    xyz
except:  # noqa: E722  # Will raise
    raise
finally:
    subscription.unregister()
```

jtbandes: Is the `except: raise` part necessary? I'm suggesting that you can avoid the `# noqa` by omitting the except clause and only using finally.

cclauss:
```
>>> try:
...     Hello
... finally:
...     print("Goodby")
...
Goodby
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
NameError: name 'Hello' is not defined
```

PR #625, opened 2021-08-20
GitHub Actions: Try pytest

cclauss:
Will https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery find our tests?


jtbandes:
The tests should be handled already by `action-ros-ci`

PR #629, opened 2021-08-21
Configure VS Code Python settings and recommended extensions

jtbandes:
cc @cclauss

Seems our CI has converged on bandit and flake8.


.gitignore
```diff
@@ -5,7 +5,12 @@
 .pydevproject
 .DS_Store
 .idea
-.vscode
+
+.vscode/*
+!.vscode/extensions.json
+!.vscode/settings.json
```
amacneil, line 11: Just remove from gitignore entirely? What else would live in this directory?

amacneil:
Suggest waiting until we run black on the whole repo and then enable formatOnSave

.gitignore
jtbandes: I have a c_cpp_properties.json in there…must’ve been automatically created? I could delete it I suppose, but it’s annoying to have it constantly show as an untracked file.

cclauss:
I would like to recommend that we use https://pre-commit.com to run our fastest lint jobs locally.

Additionally, I would like to recommend that we use https://pre-commit.ci to run those lint jobs instead of GitHub Actions.  The advantage of pre-commit.ci is that it will auto-format a pull request.

For example, see the workflow at https://github.com/kornia/kornia/pull/1210#commits-pushed-3e05ead where I make my commits and then pre-commit.ci comes in after me and automatically blackens my modifications so I don't have to.

jtbandes:
@cclauss Now that we are using pre-commit in this repo, I updated this PR to move some of the pre-commit tool configs into pyproject.toml so they can also be read when running the tools directly.

.vscode/settings.json
```diff
@@ -4,4 +4,7 @@
     "python.linting.flake8Enabled": true,
     "python.linting.banditEnabled": true,
     "python.linting.pylintEnabled": false,
-}
+    "[python]": {
+        "editor.formatOnSave": true
+    },
+}
```
cclauss: Please add one carriage return (\n).

cclauss (changes requested):

PR #623, opened 2021-08-20
Fix unused variables: flake8 --select=F841

cclauss:
Fix ~five of the~ six flake8 issues related to variables that are assigned to but then not used.
```
6     F841 local variable 'msg_id' is assigned to but never used
```

~This PR does not fix the following issue where the `--unregister_timeout` command line option should probably be removed,~  NOTE: Fixed in second commit.
```
./rosbridge_server/scripts/rosbridge_websocket.py:194:17: F841 local variable 'unregister_timeout' is assigned to but never used
                unregister_timeout = float(sys.argv[idx])
                ^
```

PR #630, opened 2021-08-21
Add pre-commit for bandit, codespell, and flake8

cclauss:
As discussed at https://github.com/RobotWebTools/rosbridge_suite/pull/629#issuecomment-903066649 add https://pre-commit.com for our fastest lint steps (bandit, codespell, flake8, and soon black and isort) to run locally in the git commit process and also in `.github/workflows/ci.yml`.

* [ ] Land this PR.
* [ ] Land the black/isort PR (after #623) and un-comment those tasks.
* [ ] ~Enable precommit.ci for this repo (autofixing) and then remove from ci.yml.~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,18 +3,17 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  lint:
+  pre-commit:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
           python-version: '3.8'
-      - run: pip install bandit codespell flake8
-      - run: bandit --recursive --skip B101,B110,B311 .
-      # See setup.cfg for the default settings for codespell and flake8
-      - run: codespell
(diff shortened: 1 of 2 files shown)


amacneil:
Open to input from other contributors, but I'm personally not convinced of the utility of this. I prefer to keep pre-commit hooks as something people can configure if they like that approach, and I'd advocate for keeping our CI all in one place (Github Actions) and limit it to running simple scripts (which can easily be run locally as well) rather than adding a dependency on another service (https://pre-commit.ci/) which will have its own downtime, potentially change or go away in the future, etc (not making any claims that its downtime is better or worse than Github Actions, just that we would now have two services that can fail).

My preferences here are because I prefer editor integration (e.g. https://github.com/RobotWebTools/rosbridge_suite/pull/629) and automatic formatting of files on save rather than pre-commit hooks or some service appending commits to PRs (shift left philosophy).

cclauss:
This PR runs pre-commit on  GitHub Actions.

I have removed the alternative CI task above to autocorrect errors.

PR #638, opened 2021-09-09
Fixed subscriber "leak" when clients disconnect

### LINKED ISSUE #580: Subscribing to a topic with a bad messageType can lead to lots of subscriptions being left open which bogs down the CPU.
If a topic is subscribed to where the "messageType" references a package that exists but has no messages, you will see the following error:

```
[Client X] [id: advertise:/foo:1] advertise: Unable to import bad_package.msg from package bad_package Caused by: No module named 'bad_package.msg'
```

This is fine, but it leaves the system in an inconsistent state. `SubscriberManager`'s `_subscribers` map never gets an entry with the topic name, but the `Subscribe(Capability)`'s `_subscriptions` map does get ont. When the client disconnects, this ends up throwing an exception where when attempting to unsubscribe this topic in the `SubscriberManager`'s unsubscribe method.

See the following stacktrace:

```
2021-08-06 13:30:35+0100 [-] Exception in thread Thread-7:
2021-08-06 13:30:35+0100 [-] Traceback (most recent call last):
2021-08-06 13:30:35+0100 [-]   File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
2021-08-06 13:30:35+0100 [-]     self.run()
2021-08-06 13:30:35+0100 [-]   File "/opt/ros/noetic/lib/python3/dist-packages/rosbridge_server/autobahn_websocket.py", line 111, in run
2021-08-06 13:30:35+0100 [-]     self.protocol.finish()
2021-08-06 13:30:35+0100 [-]   File "/opt/ros/noetic/lib/python3/dist-packages/rosbridge_library/protocol.py", line 277, in finish
2021-08-06 13:30:35+0100 [-]     capability.finish()
2021-08-06 13:30:35+0100 [-]   File "/opt/ros/noetic/lib/python3/dist-packages/rosbridge_library/capabilities/subscribe.py", line 346, in finish
2021-08-06 13:30:35+0100 [-]     subscription.unregister()
2021-08-06 13:30:35+0100 [-]   File "/opt/ros/noetic/lib/python3/dist-packages/rosbridge_library/capabilities/subscribe.py", line 88, in unregister
2021-08-06 13:30:35+0100 [-]     manager.unsubscribe(self.client_id, self.topic)
2021-08-06 13:30:35+0100 [-]   File "/opt/ros/noetic/lib/python3/dist-packages/rosbridge_library/internal/subscribers.py", line 219, in unsubscribe
2021-08-06 13:30:35+0100 [-]     self._subscribers[topic].unsubscribe(client_id)
2021-08-06 13:30:35+0100 [-] KeyError: '/foo'
```

Here's the kicker: This exception aborts the whole `finish` process, which means some subscriptions are never unsubscribed. If the client reconnects a bunch of times, this can lead to a large number of subscriptions left open, which can severely bog down the CPU.

## Expected Behavior
The system should not end up in an inconsistent state in the first place. If it does it should gracefully handle the case where the key doesn't exist in a list before accessing it. It would be better if the entire process crashed in this case.

## Actual Behavior
The unsubscribe thread throws an exception and leaves many subscriptions subscribed.

  1. Connect to rosbridge.
  2. Subscribe using a messageType referencing a package which exists but has no messages.
  3. Reconnect a bunch of times, see your CPU usage start to rise.

## Specifications

  - ROS Version: noetic
  - OS Version: focal
  - Rosbridge Version: 0.11.13
  - Twisted Version: 18.9.0

stevegolton:
Fixes #580.

Check for existence of topic key in subscribers before using it when unsubscribing subscribers when a client disconnects.
If the key doesn't exist, this fix stops the KeyError exception from propagating out of the unsubscribe loop which leave subscribers connected after clients have disconnected.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -214,11 +214,12 @@ class SubscriberManager():
 
         """
         with self._lock:
-            self._subscribers[topic].unsubscribe(client_id)
+            if topic in self._subscribers:
+                self._subscribers[topic].unsubscribe(client_id)
 
-            if not self._subscribers[topic].has_subscribers():
-                self._subscribers[topic].unregister()
-                del self._subscribers[topic]
+                if not self._subscribers[topic].has_subscribers():
+                    self._subscribers[topic].unregister()
+                    del self._subscribers[topic]
 
 
 manager = SubscriberManager()


amacneil (approved):
thanks!

MatthijsBurgh:
Why did the checks pass on this PR before merging, but after merging, they failed. Can we pinpoint the issue?

PR #657, opened 2021-09-28
Migrate remaining linters to pre-commit

amacneil:
**Public API Changes**
None

**Description**
Follow up from https://github.com/RobotWebTools/rosbridge_suite/pull/648:

- Remove separate `pre-commit.yaml` Github workflow in favor of existing `ci.yml` workflow
- Removed unused json and toml linters from `.pre-commit-config.yaml`
- `isort` is updated to v5.9.3
- `bandit`, `codespell`, and `flake8` are migrated to `.pre-commit-config.yaml` so all linters are in one place
- Fixed wildcard and invalid imports in `rosapi_node` which were missed by https://github.com/RobotWebTools/rosbridge_suite/pull/609 (due to not having a .py extension)

Related PR where this was originally suggested (fyi @cclauss): https://github.com/RobotWebTools/rosbridge_suite/pull/630

cc @kenji-miyake


rosapi/scripts/rosapi_node
```diff
@@ -277,12 +303,12 @@ class Rosapi(Node):
 def get_service_host(request):
```
amacneil, line 303: note the comment above this line indicates that this service isn't used, so I'm not sure what to make of it

cclauss (approved):
LGTM

rosapi/scripts/rosapi_node
jtbandes: Probably best to comment out or delete the code if it's not working 🤷

jtbandes (approved):

rosapi/scripts/rosapi_node
amacneil: Agreed, although I don't actually know what this code does so was erring on the safe side.

I will make a separate PR to remove it.

PR #658, opened 2021-09-28
Fix DOS line endings

amacneil:
**Public API Changes**
None

**Description**
Replace DOS line endings in several files with UNIX style.


amacneil:
@jtbandes is it safe to update these `.srv` files, or is that going to mess up something relying on md5sums?

Given their names, are they only used for testing?

chfritz:
> is it safe to update these .srv files, or is that going to mess up something relying on md5sums?

That should be fine. The md5sum is *not* computed from the file itself, but from a string that is constructed from the field names and types in a specific way that seems line ending agnostic. See, for instance, [`getMd5text()` in rosnodejs](https://github.com/RethinkRobotics-opensource/rosnodejs/blob/4aeb3c9689658609851d728b382da1b86c85f089/src/utils/messageGeneration/MessageSpec.js#L459). Meaning the md5sum should stay the same, but I haven't yet verified.

amacneil:
Great, thanks!

PR #663, opened 2021-09-30
Port #464, #478, #496, and #502 from `ros1` branch

DomenicP:
**Public API Changes**

* Synchronize `SubscriberManager` methods with a `threading.Lock`
* Add `block` parameter to `MessageHandler.finish()` method with default value of `True`
  * Use this parameter to allow `QueueMessageHandler.finish()` to run without blocking
* Add `IncomingQueue` class to `websocket_handler.py` to decouple incoming messages from the IOLoop thread

**Description**

Earlier this year I ran into some of the issues described in https://github.com/RobotWebTools/rosbridge_suite/issues/425 where the rosbridge server would end up deadlocked or blocked. I was able to solve the problem by porting several PRs from the `ros1` branch (which was `develop` at the time) to my working branch.

* https://github.com/RobotWebTools/rosbridge_suite/pull/464
* https://github.com/RobotWebTools/rosbridge_suite/pull/478
* https://github.com/RobotWebTools/rosbridge_suite/pull/496
* https://github.com/RobotWebTools/rosbridge_suite/pull/502

I recently rebased my working branch to a newer upstream version and after testing confirmed that these fixes are still needed for my application. I thought I would open a PR and see if there was any interest in getting these features into the main ROS2 branch. All credit to the original PR authors; I simply ported the changes over from the ROS1 branch to ROS2.

### DIFF
--- a/rosbridge_server/src/rosbridge_server/websocket_handler.py
+++ b/rosbridge_server/src/rosbridge_server/websocket_handler.py
@@ -34,6 +34,7 @@ import sys
 import threading
 import traceback
 import uuid
+from collections import deque
 from functools import partial, wraps
 
 from rosbridge_library.rosbridge_protocol import RosbridgeProtocol
@@ -43,6 +44,8 @@ from tornado.gen import BadYieldError, coroutine
 from tornado.ioloop import IOLoop
 from tornado.iostream import StreamClosedError
 from tornado.websocket import WebSocketClosedError, WebSocketHandler
+
+_io_loop = IOLoop.instance()
 
 
 def _log_exception():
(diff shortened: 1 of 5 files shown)


rosbridge_server/src/rosbridge_server/websocket_handler.py
```diff
@@ -65,6 +68,51 @@ def wrapper(*args, **kwargs):
...
+            with self.cond:
+                if len(self.queue) == 0 and not self._finished:
+                    self.cond.wait()
+
+                if self._finished:
+                    break
+
+                msg = self.queue.popleft()
+
+            self.protocol.incoming(msg)
```
mvollrath, line 111: A condition for `IncomingQueue` being useful is this call blocking when there is some amount of backpressure on the WebSocket.

Part of the reason the ROS1 branch switched to Autobahn/Twisted is it has explicit features to handle backpressure.  Tornado was infinitely queueing messages leading to OOM when a client was slow or unresponsive.  Tornado may have useful features for blocking when its outgoing buffer is above some threshold, but I did not find them.

mvollrath: Nevermind, was thinking of outgoing.

mvollrath (approved):

amacneil:
Thanks!

PR #667, opened 2021-10-12
Exit cleanly on SIGINT; remove sleep in test

jtbandes:
The rosbridge_websocket and rosapi nodes now exit cleanly on ctrl-C. This makes the smoke test run faster when combined with some more targeted awaits in place of a sleep.


amacneil (approved):
nice

Issue #650, opened 2021-09-25
rosbridge stops spinning after a service is called


kenji-miyake:
**Description**

Since rosbridge stops spinning, it can't receive the response [here](https://github.com/RobotWebTools/rosbridge_suite/blob/bc06a8dc911fd5ca891a5363021b03cf1f055ca4/rosbridge_library/src/rosbridge_library/capabilities/advertise_service.py#L45-L50). [`prewrite_message` function](https://github.com/RobotWebTools/rosbridge_suite/blob/bc06a8dc911fd5ca891a5363021b03cf1f055ca4/rosbridge_server/src/rosbridge_server/websocket_handler.py#L136) seems to be not running.

- Library Version: bc06a8dc911fd5ca891a5363021b03cf1f055ca4 (latest ros2 branch at 25 Sep)
- ROS Version: ROS2 Galactic
- Platform / OS: Ubuntu 20.04

**Steps To Reproduce**

Please see https://github.com/tier4/rosbridge_suite/pull/8.

**Expected Behavior**

- rosbridge can send a service call request to clients.
- rosbridge can receive the response of the service call.

**Actual Behavior**

- rosbridge can't send the request.
- rosbridge can't receive the response.

kenji-miyake:
FYI

- If I add `IOLoop.instance().start()` after [here](https://github.com/RobotWebTools/rosbridge_suite/blob/bc06a8dc911fd5ca891a5363021b03cf1f055ca4/rosbridge_server/src/rosbridge_server/websocket_handler.py#L133), it can send the request, but it of course causes another error.
- We quickly modified this [here](https://github.com/tier4/rosbridge_suite/commit/d3414b2cd6e819a4689871b3f2bc30465742df66), but I believe there is a better way to solve this, so I'd like to discuss it.

jtbandes:
I believe there is a better way to implement the service call handler. Internally, the Service system seems to support async functions being used for the callback: https://github.com/ros2/rclpy/blob/38a408867719cb76c4b4f0b1caaf9cc191336970/rclpy/rclpy/executors.py#L369 Hopefully this will allow us to remove the `while ...: sleep()` loop and maybe the lock.

jtbandes:
Here's some ideas for an alternative implementation than #652: https://github.com/RobotWebTools/rosbridge_suite/pull/666 What do you think?

kenji-miyake:
[...]

PR #669, opened 2021-10-20
Fix incoming service calls

DomenicP:
**Public API Changes**
N/A

**Description**
Not to be confused with #650 (although possibly related?).

After porting the `IncomingQueue` class from the ROS1 branch in #663, I discovered that incoming service calls from rosbridge clients were no longer working. The rosbridge server would receive the service call and run until this line in the `call_service()` function where it would get stuck:

https://github.com/RobotWebTools/rosbridge_suite/blob/181e039890ce6bf366d98dfed2828c716b3b974d/rosbridge_library/src/rosbridge_library/internal/services.py#L127

On the client side the connection would sit open ("pending" according to the Chrome debugger) waiting for a response that would never arrive.

After doing some digging I believe the problem is that we are now attempting to call `rclpy.spin_until_future_complete()` on a background thread instead of the main thread. I'm not super familiar with the details, but doing some quick research seems to indicate that at least historically trying to call `spin()` from multiple threads can cause problems. For example:

* https://answers.ros.org/question/339827/ros2-rclcpp-eloquent-spin_until_future_complete-in-an-already-spinning-node/
* https://github.com/ros2/rclpy/issues/585
* https://answers.ros.org/question/342302/ros2-rclpy-call-a-service-from-inside-a-service/

(These links are not directly related to the issue at hand but are semi-related problems I've seen regarding spinning and multiple threads.)

Since this code is running in the `IncomingQueue` thread instead of the main thread, it should be safe to use a synchronous service call here without blocking the main event loop. Making this change fixed the service calls and the client received responses again.

One question I have is if we are going to switch back to a synchronous service call, do we want to run it in the `IncomingQueue` thread? Or do we want to create a new thread for each service call so that the `IncomingQueue` is not blocked? The `ServiceCaller` class inherits from `threading.Thread` and historically under the ROS1 branch it looks like it was run with `.start()` instead of `.run()` like it is now.

Perhaps someone with a better understanding of Tornado and the event loop and threads used in rosbridge has a better idea of how this situation could be handled without adding more threads?

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/services.py
+++ b/rosbridge_library/src/rosbridge_library/internal/services.py
@@ -32,7 +32,6 @@
 
 from threading import Thread
 
-from rclpy import spin_until_future_complete
 from rclpy.expand_topic_name import expand_topic_name
 from rosbridge_library.internal.message_conversion import (
     extract_values,
@@ -123,12 +122,11 @@ def call_service(node_handle, service, args=None):
 
     client = node_handle.create_client(service_class, service)
 
-    future = client.call_async(inst)
-    spin_until_future_complete(node_handle, future)
-    if future.result() is not None:
+    result = client.call(inst)
+    if result is not None:
         # Turn the response into JSON and pass to the callback
-        json_response = extract_values(future.result())
+        json_response = extract_values(result)
     else:
-        raise Exception(future.exception())
+        raise Exception(result)
 
     return json_response
--- a/rosbridge_server/CMakeLists.txt
+++ b/rosbridge_server/CMakeLists.txt
@@ -29,6 +29,7 @@ install(FILES
 
 if(BUILD_TESTING)
   find_package(launch_testing_ament_cmake REQUIRED)
-  add_launch_test(test/websocket/smoke.test.py)
   add_launch_test(test/websocket/advertise_service.test.py)
+  add_launch_test(test/websocket/call_service.test.py)
+  add_launch_test(test/websocket/smoke.test.py)
 endif()
--- a/rosbridge_server/test/websocket/call_service.test.py
+++ b/rosbridge_server/test/websocket/call_service.test.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+import os
+import sys
+import unittest
+
+from rclpy.node import Node
+from std_srvs.srv import SetBool
+from twisted.python import log
+
+sys.path.append(os.path.dirname(__file__))  # enable importing from common.py in this directory
+
+import common  # noqa: E402
+from common import expect_messages, websocket_test  # noqa: E402
+
+log.startLogging(sys.stderr)
+
+generate_test_description = common.generate_test_description
+
+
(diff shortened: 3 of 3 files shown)


jtbandes:
Hi, thanks for your contributions! I've just refactored some of the test code (#675) to allow us to more easily write tests which use the WebSocket server. Could you please update this PR to add a test — ideally one which fails without the proposed fixes, and passes with them? Check out the tests I'm adding in #666 for an example.

jmarsik:
+1

Please fix this, rosbridge is currently unusable (at least in Foxy) with this bug :( I've spent some time debugging why Foxglove Studio tool does not display anything when connecting to our ROS2 instance through rosbridge, then I found this...

Maybe I will find some time later to write that test, but not right now.

DomenicP:
Just pushed a commit that contains a test for the service call fix. With the patch from the first commit reverted, the test stalls until it times out due to the previously described issue. With the proposed fixes the test passes successfully. I borrowed the `expect_messages()` function from `common.py` since #666 has not been merged yet.

jmarsik:
Oh, perfect, you were faster than me :) Thank you ... Now let's get it merged and released!

jtbandes:
I'll take a look at this after the holiday weekend!

jtbandes:
> historically under the ROS1 branch it looks like it was run with `.start()` instead of `.run()` like it is now.

Thanks for pointing that out, I might not have noticed it otherwise. I was going to say it looks like the service call is happening in a background thread anyway, so changing it back to synchronous is fine, but now I see it's not actually running in a separate thread anymore (and the fact that it inherits from Thread is just tech debt now).

> Perhaps someone with a better understanding of Tornado and the event loop and threads used in rosbridge has a better idea of how this situation could be handled without adding more threads?

I haven't followed all the IncomingQueue changes and am still learning this codebase, but I would generally say that no _blocking_ calls should be happening in the websocket message handler if possible. It seems rclpy provides a decent number of APIs that are async-friendly, which is how I fixed #650 without adding more threads. I wonder if we used this async approach consistently for all message handlers, perhaps we wouldn't need the IncomingQueue thread at all?

In the meantime this small fix seems useful and is something we can iterate on.

jtbandes (approved):

jtbandes:
If you enable [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) then I can help resolve the merge conflict. It's a pretty minimal conflict though.

DomenicP:
> If you enable [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) then I can help resolve the merge conflict. It's a pretty minimal conflict though.

I'd be glad to do so, except unfortunately I don't see the option available on the PR. Might be a parent org setting preventing me from enabling that option. Regardless, I'll do a quick rebase to resolve the merge conflict in just a minute.

Issue #670, opened 2021-10-26
dict_values object causes exception for new subscribers


pxmaschine:
**Description**

Please refer to

https://github.com/RobotWebTools/rosbridge_suite/blob/cc5209cfcc2f9732c20f7741f1e029430b69f12c/rosbridge_library/src/rosbridge_library/internal/subscribers.py#L172-L195

as well as

https://github.com/RobotWebTools/rosbridge_suite/blob/cc5209cfcc2f9732c20f7741f1e029430b69f12c/rosbridge_library/src/rosbridge_library/internal/subscribers.py#L209

In line 209, a list of a [dictionary view](https://docs.python.org/3/library/stdtypes.html#dict-views) is created (using Python 3.X interpreter; Python 2.X interpreter should result in a list of list type).  Line 190 then iterates over said list that contains only one object of type `dict_values`. A call on the dict_values object is then executed in line 192, which results in the following Exception:

`'dict_values' object is not callable`

- Library Version: 605e814
- ROS Version: ROS_VERSION=2  ROS_DISTRO=galactic
- Platform / OS: Ubuntu 20.04.3 LTS
- Python Version: 3.8.10

**Steps To Reproduce**

Any way to force execution of line 197 should result in the same exception.

Our specific case:
Behaviour can be forced when hard resetting an MCU running a websocket client (e.g. ESP8266 + [WebSockets](https://platformio.org/lib/show/549/WebSockets/)) that subscribes to some topic. This will cause a client duplication on the rosbridge_server, which will only get cleaned up after the first published message. When a message is published for the first time to the subscribed topic, this will cause a call to
https://github.com/RobotWebTools/rosbridge_suite/blob/cc5209cfcc2f9732c20f7741f1e029430b69f12c/rosbridge_library/src/rosbridge_library/internal/subscribers.py#L197
and results in said exception.

**Expected Behavior**
Creation of list of subscription callables in line 209.

This can be achieved by changing line 209 from

`            self.callback(msg, [self.new_subscriptions.values()])`

to

`            self.callback(msg, self.new_subscriptions.values())`

**Actual Behavior**
Creation of list of dict_values of subscription callables in line 209.

jtbandes:
I've added a test to show that this fixes the bug: https://github.com/RobotWebTools/rosbridge_suite/pull/687

jtbandes:
Thanks for the report and fix!

PR #671, opened 2021-10-26
Fix for #670

pxmaschine:
Fix for #670

**Public API Changes**

None

**Description**

Fixes apparent typing error: List[List[callable]] -> List[callable] 

#670

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -206,7 +206,7 @@ class MultiSubscriber:
         subscriptors.
         """
         with self.lock:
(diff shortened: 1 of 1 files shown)


pxmaschine:
This error was introduced by 8dea92f2285c18db35f58ace625e42ec66bb0437. The commit shows no indication for using a nested list in this case. This seems like a typical typo to me. Please fix this, as this is a [long standing issue](https://github.com/RobotWebTools/rosbridge_suite/pull/562) with the core library of rosbridge!

eborghi10:
Hey guys, we're also experiencing issues and this fix will help us a lot. We can provide a failure in case you need to replicate it.

CC @DiegoHollmann

Issue #515, opened 2020-08-02
Exception in deserialization of BSON with ROS Noetic


repa1030:
##  Expected Behavior
ROSBRIDGE Server works fine with ROSSharp and BSON serialization such as in ROS Melodic.

## Actual Behavior
Getting Error:
[Client 0] Exception in deserialization of BSON

## Steps to Reproduce the Problem
  1. Use current ROSBRIDGE develop branch
  2. Use current ROSSharp branch
  3. Send any message with BSON serialization

## Specifications

  - ROS Version (`echo $ROS_DISTRO`): Noetic
  - OS Version (`grep DISTRIB_CODENAME /etc/lsb-release`): Focal
  - Rosbridge Version (`roscat rosbridge_server package.xml | grep '<version>'`): 0.11.9
  - Twisted Version (`python -c 'import twisted; print twisted.version'`): 18.9.0

RyodoTanaka:
I have the same error, does any one have the solution ... ?

nick-paul:
#545 Fixes some major python3/BSON issues. Does using that branch resolve the issue?

repa1030:
Yes, fixed :)

PR #545, opened 2020-12-14
Fix bson support for python3

nick-paul:
When running in BSON only mode with python3, the bridge silently fails and does not publish any messages. 

An exception is caught on [line 109 of tcp_handler.py](https://github.com/RobotWebTools/rosbridge_suite/blob/a09a964fb5956321aca3b296da367e21d3d2e044/rosbridge_server/src/rosbridge_server/tcp_handler.py#L109) and ignored. Re-raising the exception gives the following error:

```
[INFO] [1607978704.270194]: Rosbridge TCP server started on port 9090
[INFO] [1607978704.671599]: [Client 0] connected. 1 client total.
[INFO] [1607978704.672726]: [Client 0] disconnected. 0 client total.
----------------------------------------
Exception happened during processing of request from ('10.0.0.14', 53436)
Traceback (most recent call last):
  File "/usr/lib/python3.8/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.8/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.8/socketserver.py", line 720, in __init__
    self.handle()
  File "/home/nick/ws/src/isolate/rosbridge_suite/rosbridge_server/src/rosbridge_server/tcp_handler.py", line 110, in handle
    raise e
  File "/home/nick/ws/src/isolate/rosbridge_suite/rosbridge_server/src/rosbridge_server/tcp_handler.py", line 96, in handle
    if self.recv_bson() == None:
  File "/home/nick/ws/src/isolate/rosbridge_suite/rosbridge_server/src/rosbridge_server/tcp_handler.py", line 68, in recv_bson
    raw_msglen = self.recvall(BSON_LENGTH_IN_BYTES)
  File "/home/nick/ws/src/isolate/rosbridge_suite/rosbridge_server/src/rosbridge_server/tcp_handler.py", line 62, in recvall
    data += packet.decode()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc9 in position 0: invalid continuation byte
----------------------------------------
```

[...]

PR #672, opened 2021-10-26
Fix BSON Deserialization Error One Failure Kill

philip-klein:
**Public API Changes**
None

**Description**
If Deserialization has an exception, there are cases in the logic tree that result in the buffer not getting cleared. This means you will spin forever attempting to deserialize a broken message and lock up the entire system.

By clearing the buffer in BSON mode when the exception is thrown, a deadlock condition can be avoided.

work-around for #515

### DIFF
--- a/rosbridge_library/src/rosbridge_library/protocol.py
+++ b/rosbridge_library/src/rosbridge_library/protocol.py
@@ -127,7 +127,6 @@ class Protocol:
         # ..first, try to load the whole buffer as a JSON-object
         try:
             msg = self.deserialize(self.buffer)
-            self.buffer = ""
 
         # if loading whole object fails try to load part of it (from first opening bracket "{" to next closing bracket "}"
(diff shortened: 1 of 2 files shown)


nick-paul:
#545 also fixes the buffer issue. It doesn't include the test modification made in this PR though. Perhaps they are also fixed in #545

amacneil:
@philip-klein @nick-paul now that we have merged #545, could one of you take a look at whether we can rebase this PR to include just that fixed unit test?

philip-klein:
Actually, now that #545 is merged, I think it makes more sense to close this MR. I think the underlying issue was actually fixed by #545 too. I'll take a look at it anyway just to make sure.

amacneil:
Sounds good. It might be worth looking into that test you fixed (it might be fixed now), where it said "Next two lines should be removed when this is fixed". It might be fixed now. I don't know this codebase well though.

Issue #673, opened 2021-10-27
the "id" field in messages is not used for subscriptions


dogoepp:
When subscribing to topics from a websocket client, I would like that the data in the ID field in my subscription operation be used in subsequent messages from the ros topic.

For instance, if I send

```json
{
  "op": "subscribe",
  "topic": "/text",
  "id": "subscription-to-topic-text-for-the-sake-of-good"
}
```

I would like that messages contain the same ID field.

```json
{
  "op": "publish",
  "topic": "/text"
  "id": "subscription-to-topic-text-for-the-sake-of-good"
  "msg": {
      "data": "Here we go"
  },
}
```

Since it seemed rather simple to implement, I am preparing a Pull Request for that.

6-000-000:
Any updates on this? Looking for this feature as well. Will this work on ROS2 if ever?

PR #674, opened 2021-10-27
Add support for the "id" field in subscribe.

dogoepp:
**Public API Changes**

Implement the protocol's documentation regarding the use of the optional "id" field for the "subscribe" and "unsubscribe" operations.

**Description**

For the "subscribe" and "unsubscribe" operations, the "id" field is now
taken into account. This means that when it is provided, the subsequent
"publish" operations, from the bridge to the client, will carry the "id"
field with the same value.

Also, if multiple subscriptions are done on the same topic with
different values for "id", the related messages are sent as many times
as there are subscriptions with different "id" values.

If one uses the "unsubscribe" operation on a topic without the "id"
field, all subscriptions to that topic are closed.

** Related Issue **

#673

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/subscribe.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/subscribe.py
@@ -151,7 +151,7 @@ class Subscription():
     def _publish(self, message):
         """ Internal method to propagate published messages to the registered
         publish callback """
-        self.publish(message, self.fragment_size, self.compression)
+        self.publish(message, self.fragment_size, self.compression, self.clients.keys())
 
     def on_msg(self, msg):
         """ Raw callback called by subscription manager for all incoming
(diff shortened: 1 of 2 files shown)


amacneil:
Hi, thanks for the PR!

Can you take a look at adding a unit test for this feature? After that, we're happy to merge.

dogoepp:
I am working on a unit test but now and then I fail to run the unit tests already there with the commands `rostest rosbridge_library test_capabilities.test` or `catkin_make run_tests`. They hang every now and then. Am I doing something wrong or should I open an issue ?

Apologies for the naive question, I am (unfortunately still) new to unit tests.

Paralelly, I just commited and sent the unit tests I wrote. I am looking forward to your comment on them.

defunctzombie:
changes LGTM - @dogoepp can you look into the failing melodic and noetic CI and see if thats related to this change?

dogoepp:
I would like to, but the CI logs seem to have been erased. Is there a way to run the checks again?

jtbandes:
I don't see a re-run button...I would recommend just pushing another commit (you can make an empty commit with `--allow-empty`)

dogoepp:
It seems that we cannot outsmart Github. It did not trigger the CI again.

jtbandes:
I do see logs for the failures now—you can’t see them?

defunctzombie:
CI is failing because the tests are failing.

dogoepp:
I can see logs now. I'll give it a look as soon as possible.

PR #675, opened 2021-11-05
Refactor smoke test to make WebSocket test harness reusable

jtbandes:
**Public API Changes**
None

**Description**
Factor out the ROS and Twisted setup code from the smoke test, so we can reuse it for other tests in the future.

### DIFF
--- a/rosbridge_server/test/websocket/smoke.test.py
+++ b/rosbridge_server/test/websocket/smoke.test.py
@@ -37,6 +37,7 @@ class TestWebsocketSmoke(unittest.TestCase):
         ws_completed_future = rclpy.task.Future()
 
         def sub_handler(msg):
+            node.get_logger().info(f"Received message via ROS subscriber: {msg}")
             ros_received.append(msg)
             if len(ros_received) == NUM_MSGS:
                 node.get_logger().info("Received all messages on ROS subscriber")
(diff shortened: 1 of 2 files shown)


rosbridge_server/test/websocket/smoke.test.py
```diff
@@ -1,215 +1,101 @@
...
-from autobahn.twisted.websocket import WebSocketClientFactory, WebSocketClientProtocol
-from rcl_interfaces.srv import GetParameters
-from rclpy.executors import SingleThreadedExecutor
 from std_msgs.msg import String
-from twisted.internet import reactor
-from twisted.internet.endpoints import TCP4ClientEndpoint
 from twisted.python import log
 
-log.startLogging(sys.stderr)
+sys.path.append(os.path.dirname(__file__))  # enable importing from common.py in this directory
```
jtbandes, line 11: There might be a better way to make this import work but I wasn't aware of one. For comparison, the rosbag2 tests use `APPEND_ENV PYTHONPATH=${CMAKE_CURRENT_SOURCE_DIR}` in their CMake rules, but we're using `add_launch_test()` in this package which doesn't seem to support the same flags. cc @emersonknapp 

https://github.com/ros2/rosbag2/blob/ccc0a9ccd7991348f5b91b29f7f267336f17016f/rosbag2_py/CMakeLists.txt#L163-L168

Issue #569, opened 2021-05-12
Fine-granular access control


uahic:
Rosauth works but is a binary access control which either gives you access to all or none 'ressources' accessible via the bridge.
I know that ROS 1 was not really intended to be used in production systems but we do have an industrial project that relies on ROS and the client requested different levels of access control. 
Currently I see two ways
1) creating multiple bridges, each with different filters on topics/services/... (let's call them 'ressources') and request authorization from a webserver, which replies with a token that is accepted by the bridge, corresponding to a particular access-level (e.g. 'admin')

2) relay restricted ressources through a custom webserver (thereby actually creating a second kind of ros-bridge for the poor man)

None of these alternative solutions is really 'clean' 
An implementation in the bridge would probably require appending the client some sort of authentication (token,... like JWT?), and needs to be checked on each request. It does not to query each time another backend server. The bridge just needs to know:  1) the mapping level <-> ressources  2) authenticate the user

Importantly, it should also be possible for 'public' users to use non-restricted topics/services

Finally, I would like to ask you for your opinions. Possibly I could do an initial implementation if people do not disapprove with this feature

## Specifications

  - ROS Version noetic
  - OS Version Ubuntu 20.04
  - Rosbridge Version - most recent

reinzor:
Hi @uahic , we encountered the same problem you wrote up here. We chose the first solution having a `client` bridge and an `admin` bridge and separate the allowed data streams via the configured `globs`.

I agree that it would be interesting to have a more generic set-up for authentication however this would have quite some impact on how things are handled in the bridge side but definitely worth some thought. I think we could still use the original [rosauth](https://github.com/GT-RAIL/rosauth/blob/develop/srv/Authentication.srv) service interface with a different implementation, e.g. JWT authorization. It already includes some `level` setup.

@mvollrath what do you think?

zflat:
A relatively simple approach would include changes to rosauth and rosbridge.

# rosauth

* Allow configuring with multiple secret files based on user "level"
* When an authentication request is made, read the user "level" from the request to determine the secret to use

# rosbridge

* When the client is authenticated, store the user "level" in addition to just True/False authenticated
* Modify the `topics_glob` (and other similar globs) so that they are scoped to specified user "levels"
* Modify the `topics_glob` to have more granularity for read vs. write. So give the option to specify `topics_read_glob`, `topics_write_glob` as separate lists
  * Use the `topics_read_glob` when subscribing
[...]

PR #676, opened 2021-11-11
Add Authentication and Authorization to ros2

zflat:
## Description ##

This is a pull request into the `ros2` branch that provides authentication and authorization capabilities via configurable service names.

When running the `rosbridge_webserver rosbridge_websocket.py` websocket server it will be possible to specify `authentication_service` and `authorization_service` service names.

When the `authentication_service` is provided, the functionality removed in https://github.com/RobotWebTools/rosbridge_suite/pull/586 is restored but the authentication service uses an updated and more generic `rosbridge_msgs/srv/Authentication.srv` service description. Of note, the connection's "client_id" is shared with the authentication service so that it can be associated with the authentication request and used for authorization.

When the `authorization_service` is provided, relevant ROS operations (like publish, advertise, etc.) are authorized against the provided service with the connection's "client_id". This allows for finer grain control that is application specific.
  * If the authorize service needs to provide access levels where topics are publishable for authenticated users of sufficient permissions, the "client_id" sent in the original authentication request can be associated with the authentication credential's permissions. Then, the client_id's permissions level can be checked to determine if publishing is authorized. 
  * It is also possible to configure authorization without authentication. For example, an authorize service that only allows subscribe operations (returns authorized=false for all other ops) effectively makes topics read-only for all connections that are authorized with that service. A simple authorize service like that could be used without authentication.

This PR only provides authentication for the websocket protocol. Further work could be done to allow authorization via a service with the "auth" protocol operation to bring authentication capabilities back to the other protocols.

### rosbridge_msgs ##

New service descriptions are added to rosbridge_msgs

- rosbridge_msgs/srv/Authorization.srv
- rosbridge_msgs/srv/Authentication.srv

### rosbridge_websocket script ###

Two new command line arguments/rosparams are added to the rosbridge_websocket script:
- authentication_service
- authorization_service

## Relevant Issues ##

- As of https://github.com/RobotWebTools/rosbridge_suite/pull/586 authentication features were removed from `ros2` in order to remove the dependency on rosauth.
- Here is discussion about finer grain access control that helped motivate this work: https://github.com/RobotWebTools/rosbridge_suite/issues/569


amacneil:
Hi @zflat

Thanks for submitting this PR. We discussed it at the [ROS web working group](https://docs.google.com/document/d/1esrRj2x80TSCEHOwuh-cT_bQvAqqvloQAQr8mPMzcTc/edit) today.

Unfortunately, rosbridge_suite is lacking active maintainers, and those who are maintaining it are working part time on it. So, we are hesitant to introduce new functionality without corresponding increase in people who can maintain the project long term.

If you're using rosbridge in production, we'd love to chat - please feel free to join one of our biweekly meetings (link in doc above). Otherwise, we don't feel that we can introduce new functionality right now, and are considering rosbridge_suite "feature complete" (bugfixes only). 

I'm going to close this PR for now. If you are interested in getting more actively involved, please come to a working group meeting and we can chat further about it.

zflat:
@amacneil Thank you for looking into this feature. I think that rosbridge is unique compared to other ROS packages because it acts as a gateway between systems and is the perfect place to introduce a layer of authentication and authorization. If authentication and authorization remain on the roadmap I just ask that it is implemented in an extensible manner so that 3rd party authentication and authorization systems can be used.

amacneil:
Sounds good. Agreed - it would be a good feature to add eventually.

PR #679, opened 2021-11-21
(server) remove execution from setup.py

MatthijsBurgh:
**Public API Changes**
None

**Description**
`setup.py` doesn't need to be executable.

P.S. file is not empty, there are just no changes in the file. Only in execution permissions


amacneil (approved):

jtbandes:
What’s the motivation for changing this? If we want to remove the executable bit we should probably also remove the `#!/usr/bin/env python`.

MatthijsBurgh:
@jtbandes I experienced that `setuo.oy` would show up in roarun. That triggered me to check a few different other pure python and ros python packagss. All didn't have execution permissions. Though most of them had a shebang line.

MatthijsBurgh:
CI is not failing because of this PR. Trying to fix CI in #680, which is WIP at the moment.

amacneil:
@MatthijsBurgh seems common in some official ROS packages to remove shebang:

https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosmsg/setup.py

I don't feel strongly about it, but it seems pointless having a shebang if the exec bit is not set?

PR #680, opened 2021-11-22
Fix ROS1 CI

### LINKED ISSUE #682: ROS 1 CI Issue
CI in Noetic is broken.

Last run that succeeded is from augustus 2021. Since then changes have been made to rospy. The issue is caused by the changes in https://github.com/ros/ros_comm/commit/b3f8e3c757cb2da35756c95bc16459a372e7874d.

More especially:
https://github.com/ros/ros_comm/commit/b3f8e3c757cb2da35756c95bc16459a372e7874d#diff-2657ef9f5903b672eeb4e9537e18bdc75de0542f6703cc160e184a889a077ff5R810-R814

This results in the process breaking out of the while receive loop. Therefore closing the subscriber. Before this commit, it would stay in the loop, as it was just catched in the general `except`. 
(I can't figure out how it would exit the loop before the commit. As all my machines already have the newest `rospy` version.)

I suggest to change the test so that is succeeds in Noetic. And either request a backport to Melodic or skip the test in Melodic.

MatthijsBurgh:
**Public API Changes**
None

**Description**
Behavior or `rospy` in noetic has changed. Therefore test needs to be changed.

Also `setuptools` has changed. It now forces the use of their `distutils` This breaks catkin stuff. Therefore we need to set the environment variable in CI. To use the stdlib version.

Fixes #682

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,5 +47,7 @@ jobs:
           required-ros-distributions: ${{ matrix.ros_distribution }}
 
       - uses: ros-tooling/action-ros-ci@v0.2
+        env:
+          SETUPTOOLS_USE_DISTUTILS: stdlib
         with:
           target-ros1-distro: ${{ matrix.ros_distribution }}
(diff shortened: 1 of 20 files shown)


amacneil (approved):
awesome work, thanks!

PR #689, opened 2021-12-07
Allow subscribing to any qos profile when creating a subscriber

defunctzombie:
**Public API Changes**

None

**Description**

When looking into https://github.com/foxglove/studio/issues/1987 I discovered that rosbridge was not receiving "latched" `/tf_static` messages. This led us to learn that subscribers need specific qos profile settings for certain publishers (https://github.com/ros2/ros2cli/issues/523#issuecomment-987107534).

This PR attempts to use a qos profile that would accept any published message. I suspect what I've done is not actually correct. There may not be such a generic qos profile and we need to be more explicit based on the publisher. I suspect the real fix is probably one of these other PRs but if there is a generic qos profile that works for any publisher that would be nice.

https://github.com/RobotWebTools/rosbridge_suite/pull/654
https://github.com/RobotWebTools/rosbridge_suite/pull/582

/cc @kenji-miyake @MrBlenny

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -33,6 +33,8 @@
 
 from threading import Lock
 
+from rclpy.qos import QoSDurabilityPolicy, QoSProfile
+
 from rosbridge_library.internal import ros_loader
 from rosbridge_library.internal.message_conversion import msg_class_type_repr
 from rosbridge_library.internal.outgoing_message import OutgoingMessage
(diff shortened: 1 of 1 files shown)


rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -107,9 +109,16 @@ def __init__(self, topic, client_id, callback, node_handle, msg_type=None, raw=F
...
         self.msg_class = msg_class
         self.node_handle = node_handle
-        # TODO(@jubeira): add support for other QoS.
+
+        # Some publishers (like /tf_static) publish messages with qos profiles that are incompatible
+        # with the default qos profile. This qos profile indicates we are ok receiving any message types.
+        qos = QoSProfile(
+            depth=10,
+            durability=QoSDurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
+        )
```
jhurliman, line 118: The main thing missing is switching between reliable and best-effort based on what is being published. Attempting to open a reliable subscription to a best-effort publisher will fail while opening a best-effort subscription on a reliable publisher could unnecessarily drop messages.

defunctzombie: Should we use https://github.com/RobotWebTools/rosbridge_suite/pull/654 instead which attempts to mirror the qos profile of the publisher?

rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -107,9 +109,16 @@ def __init__(self, topic, client_id, callback, node_handle, msg_type=None, raw=F
...
+        node_handle.get_logger().info(topic)
```
defunctzombie, line 119: fixme - remove

defunctzombie:
Closing in favor of https://github.com/RobotWebTools/rosbridge_suite/pull/690

PR #690, opened 2021-12-07
Allow subscribing to any qos profile when creating a subscriber

### LINKED ISSUE #551: QoS compatibility failing silently (ROS2)

defunctzombie:
**Public API Changes**
Subscriber QoS is automatically chosen based on available publishers to achieve compatible QoS settings in more cases.

**Description**
When looking into https://github.com/foxglove/studio/issues/1987 I discovered that rosbridge was not receiving "latched" `/tf_static` messages. This led us to learn that subscribers need specific qos profile settings for certain publishers (https://github.com/ros2/ros2cli/issues/523#issuecomment-987107534).

This PR attempts to use a qos profile that would accept any published message.

Closes https://github.com/RobotWebTools/rosbridge_suite/pull/654
Closes https://github.com/RobotWebTools/rosbridge_suite/pull/582
Fixes #551 

/cc @kenji-miyake @MrBlenny @DomenicP


defunctzombie:
👍

jtbandes (approved):

PR #706, opened 2022-01-28
reorganize into smaller methods

zflat:
This cleans up code style. The main change is to the `RosbridgeWebsocketNode` class in `rosbridge_websocket.py` to fix linting errors about methods that are too many lines.

**Public API Changes**
Removes the `authenticate` ros param since the code using that param has been deleted.

**Description**
These are code cleanup changes that I implemented in https://github.com/RobotWebTools/rosbridge_suite/pull/676 but since that was closed I am making a smaller PR that is purely code cleanup.


jtbandes (approved):
Looks like a relatively straightforward reorganization, thanks! Sorry for the delayed review.

PR #707, opened 2022-01-28
Explicitly rename param subnode

### LINKED ISSUE #704: Websocket roslaunch launches rosapi twice
**Description**

Hi, I have stumbled across an odd bug where when I try to launch `ros2 launch rosbridge_server rosbridge_websocket_launch.xml`, it launches two rosnodes named `rosapi` and thus throws a warning. It hasn't caused any problems until now where (I think) it is interferring with my connection to Foxglove:

Launch file:
```
root@f0667d79a4bd:/# ros2 launch rosbridge_server rosbridge_websocket_launch.xml 
[INFO] [launch]: All log files can be found below /root/.ros/log/2022-01-20-11-04-39-032547-f0667d79a4bd-40
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [rosbridge_websocket-1]: process started with pid [42]
[INFO] [rosapi_node-2]: process started with pid [44]
[rosapi_node-2] [WARN] [1642676679.322604700] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
[rosbridge_websocket-1] [INFO] [1642676679.338517200] [rosbridge_websocket]: Rosbridge WebSocket server started on port 9090
[rosbridge_websocket-1] [INFO] [1642676680.439530400] [rosbridge_websocket]: Client connected. 1 clients total.
```
Checking nodes:
```
root@7c665ff25d17:/ros_ws# ros2 node list
WARNING: Be aware that are nodes in the graph that share an exact name, this can have unintended side effects.
/rosapi
/rosapi
/rosbridge_websocket
```

The only place I've found reference to this is in this old post: https://answers.ros.org/question/344141/ros2-launch-creates-two-nodes-of-same-type/ and also this issue: https://github.com/ros2/launch/issues/411 which links to this issue https://github.com/ros2/geometry2/issues/124 

As mentioned in those posts, removing the `name=rosapi` field of the rosapi node entry within the launchfile appears to solve this problem (locally for me). 

i.e. from:
```
 <node name="rosapi" pkg="rosapi" exec="rosapi_node">
    <param name="topics_glob" value="$(var topics_glob)"/>
    <param name="services_glob" value="$(var services_glob)"/>
    <param name="params_glob" value="$(var params_glob)"/>
  </node>
```
to
```
  <node pkg="rosapi" exec="rosapi_node">
    <param name="topics_glob" value="$(var topics_glob)"/>
    <param name="services_glob" value="$(var services_glob)"/>
    <param name="params_glob" value="$(var params_glob)"/>
  </node>
```

Hopefully this isn't just me, but apologies if I've done something wrong in the setup! 

- Library Version: latest? not sure how to check! 
- ROS Version: Foxy
- Platform / OS: Ubuntu 20.04 (via latest docker container)

**Steps To Reproduce**

Running
```
ros2 launch rosbridge_server rosbridge_websocket_launch.xml
```
then checking the active nodes using `ros2 node list`

**Expected Behavior**
Ros launching the file will start a single rosapi node and the websocket node, I guess it should show as creating two nodes, `rosapi` and `rosapi_params`

**Actual Behavior**
Ros launching thefile creates two nodes named `rosapi` and ros therefore throws warnings.

rob-clarke:
**Public API Changes**
None(ish)

The param subnode is now named `{parent_node_name}_params` when the rosapi node is renamed. Previously it would have the same name.

**Description**
Previously, without any renaming, running the `rosapi` node would spawn a `rosapi_params` subnode. If the top-level node was renamed (or had a name set by a launch file) the param subnode ended up with the same name as the top-level node. This fixes the behaviour by explicitly setting the name of the subnode to the same as the top-level node with `_params` as a suffix.

Unlike the solution described in [the issue](https://github.com/RobotWebTools/rosbridge_suite/issues/704), this preserves any other global arguments. This means when changing namespace of the top-level node, the subnode will have a matching namespace.

Fixes #704

### DIFF
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -70,7 +70,11 @@ def init(parent_node_name):
     global _node, _parent_node_name
     # TODO(@jubeira): remove this node; use rosapi node with MultiThreadedExecutor or
     # async / await to prevent the service calls from blocking.
-    _node = rclpy.create_node("rosapi_params")
+    parent_node_basename = parent_node_name.split("/")[-1]
+    param_node_name = f"{parent_node_basename}_params"
+    _node = rclpy.create_node(
+        param_node_name, cli_args=["--ros-args", "-r", f"__node:={param_node_name}"]
+    )
     _parent_node_name = get_absolute_node_name(parent_node_name)
 
 


jtbandes (approved):

jtbandes:
Thanks!

PR #709, opened 2022-02-02
Feature config url_path for deployment

sirawats:
**Public API Changes**

None

**Description**

1. Fix args `--address` type bug. replaced type int() with str()
2. Add new args `--url_path` and ROS param

url_path should be a configurable parameter because when application is deployed the port (9090) need an ingress to url path.
(for e.g. 127.0.0.0:9090 -> xxx.ap-southeast-1.elb.amazonaws.com/rosbridge_websocket)

https://kubernetes.io/docs/concepts/services-networking/ingress/

### DIFF
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -146,12 +146,11 @@ class RosbridgeWebsocketNode(Node):
         # Done with parameter handling                   #
         ##################################################
 
+        handlers = [(r"/", RosbridgeWebSocket), (r"", RosbridgeWebSocket)]
         if url_path != "/":
-            application = Application([(rf"{url_path}", RosbridgeWebSocket)], **tornado_settings)
-        else:
-            application = Application(
-                [(r"/", RosbridgeWebSocket), (r"", RosbridgeWebSocket)], **tornado_settings
-            )
+            handlers = [(rf"{url_path}", RosbridgeWebSocket)]
+
+        application = Application(handlers, **tornado_settings)
 
         connected = False
         while not connected and self.context.ok():
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -124,6 +124,15 @@ class RosbridgeWebsocketNode(Node):
                 print("--address argument provided without a value.")
                 sys.exit(-1)
 
+        url_path = self.declare_parameter("url_path", "/").value
+        if "--url_path" in sys.argv:
+            idx = sys.argv.index("--url_path") + 1
+            if idx < len(sys.argv):
+                url_path = str(sys.argv[idx])
+            else:
+                print("--url_path argument provided without a value.")
+                sys.exit(-1)
+
         retry_startup_delay = self.declare_parameter("retry_startup_delay", 2.0).value  # seconds.
         if "--retry_startup_delay" in sys.argv:
             idx = sys.argv.index("--retry_startup_delay") + 1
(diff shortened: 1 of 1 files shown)


zflat:
@leossok @jtbandes The linting that is failing here (https://github.com/RobotWebTools/rosbridge_suite/runs/5040354094?check_suite_focus=true) is exactly why I opened the PR to refactor in https://github.com/RobotWebTools/rosbridge_suite/pull/706

amacneil:
@leossok can you please rebase this PR to fix merge conflicts?

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -155,6 +223,8 @@ def __init__(self):
             if len(element.strip().strip("'")) > 0
         ]
 
+<<<<<<< HEAD
+=======
```
zflat: Source control conflict markers and a bunch of other commits showing up in this pull request. I think you should try to fix the rebase.

sirawats: Oh, my mistake. Now I think I resolved those conflict.

amacneil:
FYI rolling is still failing CI. Let us know when you get it green and someone can review the PR.

If you're stuck on why it won't pass tests, please let us know and someone can try to help.

MatthijsBurgh:
@amacneil The CI is also failing on the ROS2 branch. It has been for more than a month. It isn't this PR. But you should fix the CI before merging any new stuff.

amacneil:
Ah, I did not notice that. Yes, we should fix that before merging further PRs.

MatthijsBurgh:
Cause of broken CI: https://github.com/RobotWebTools/rosbridge_suite/issues/725

We should ignore it for now.

defunctzombie:
It looks like there are a number of unrelated changes in this PR (maybe a bad merge or rebase?). I think it would be better to consider those changes separately. @leossok can you take a look at the current diff and see what is related to this feature and what is not. That would help with PR review.

sirawats:
Sorry for late fix, please review.

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -137,9 +146,12 @@ def __init__(self):
...
 
-        application = Application(
-            [(r"/", RosbridgeWebSocket), (r"", RosbridgeWebSocket)], **tornado_settings
-        )
+        if url_path != "/":
+            application = Application([(rf"{url_path}", RosbridgeWebSocket)], **tornado_settings)
+        else:
+            application = Application(
+                [(r"/", RosbridgeWebSocket), (r"", RosbridgeWebSocket)], **tornado_settings
+            )
```
MatthijsBurgh: I would go for defining the list in an if/else and define the application afterwards.

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -124,6 +124,15 @@ def __init__(self):
...
                 sys.exit(-1)
 
+        url_path = self.declare_parameter("url_path", "/").value
+        if "--url_path" in sys.argv:
+            idx = sys.argv.index("--url_path") + 1
+            if idx < len(sys.argv):
+                url_path = str(sys.argv[idx])
+            else:
+                print("--url_path argument provided without a value.")
+                sys.exit(-1)
```
MatthijsBurgh, line 134: I see that this is copied from above. Though I think this code is sensitive for the following error:
`--option1 --option2 value2`.
So it might be good to change this in another PR.

defunctzombie: Is there an accepted and typical args parsing library that python projects use or should be using? I'd say its outside the scope of this PR to switch but it does seem like it would avoid hand-crafted args parsing like what is happening here.

defunctzombie:
@MatthijsBurgh is this PR 👍 from your perspective?

rosbridge_server/scripts/rosbridge_websocket.py
zflat: https://github.com/docopt/docopt looks pretty nice. I wonder how hard it would be to use it while still keeping the CLI backwards compatible with the current implementation.

MatthijsBurgh: How about the standard [argparse module](https://docs.python.org/3/library/argparse.html)?

MatthijsBurgh (approved):

PR #710, opened 2022-02-05
Fix formatting of operations list

rgov:
**Public API Changes**
None

**Description**
Fixes what I suspect is a mistake in the spec, where `publish` is a separate operation.


jtbandes (approved):
Thanks!

PR #711, opened 2022-02-06
Exapnding CBOR value extraction support

kedus42:
**Public API Changes**
None

**Description**
Changes allow CBOR value extraction to now also support messages with no __slot__ or _slot_types attributes.


jtbandes:
Thanks for your contribution. Can you please provide some background explanation on this change? Why would any messages not have `__slots__` or `_slot_types`? Can you provide examples in the form of unit tests?

mvollrath:
Looks like there is also a feature for rosapi providing ROS version in this PR, maybe an accident.

amacneil:
@kedus42 can you please rebase this PR to remove the extra commits?

defunctzombie:
Close since we've not heard back about this. If you are able to revisit we can take another look.

PR #712, opened 2022-02-10
Remove authentication, link to ROS 1 spec

amacneil:
**Public API Changes**
None

**Description**
- Authentication is not supported in ROS 2 rosbridge
- Add link to ros1 branch


jtbandes (approved):
Should we just ditch the numbering entirely?

amacneil:
Probably. Or if the protocol is going to be versioned, then it should specify as much in the subprotocol?

jtbandes:
I mostly meant the section numbering in the markdown.

amacneil:
Oh, yeah could do. Maybe in a separate PR? If we did that the it would probably be best to addd a ToC.

I don't think the spec will evolve much tho? Will leave as is for now.

PR #723, opened 2022-03-12
Fixed multiple subscriber on transient_local topic

p0rys:
**Public API Changes**
None

**Description**
Problem: Only the first subscription to a transient_local topic does receive the "latched" data.

The MultiSubscriber class creates a new subscription to initially receive the "latched" data for each new client. After this subscription returns with a message the client is shifted over to the "real" subscription and the other one is destroyed. Because this intermittent subscribers does not use the same QoS settings as the real one it will never receive latched data from transient_local topics except there is send new data from the publisher.

Fix: This PR adds the QoS settings as a global variable to the MultiSubscriber class and uses it when creating intermittent subscribers for new clients. With the correct settings these will directly get back with the latched data. 

The transient_local_publisher test is also changed to subscribe three times to cover this case.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -101,21 +101,6 @@ class MultiSubscriber:
         if topic_type is not None and topic_type != msg_type_string:
             raise TypeConflictException(topic, topic_type, msg_type_string)
 
-        # Create the subscriber and associated member variables
-        # Subscriptions is initialized with the current client to start with.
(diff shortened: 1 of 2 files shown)


p0rys:
Rolling CI see #725

defunctzombie:
@mvollrath @MatthijsBurgh does this change looks reasonable to you? It looks 👍 to me.

MatthijsBurgh (approved):
I am not familiar with ROS2. So I can only judge the syntax, which looks fine.

jtbandes (approved):

PR #724, opened 2022-03-15
(dependabot) ignore patch updates

MatthijsBurgh:
Dependabot should ignore all patch updates.

It will only create PRs for minor and major updates. Security updates are not affected.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,5 +5,5 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      dependency-name: "*"
-      update-types: ["version-update:semver-patch"]
+      - dependency-name: "*"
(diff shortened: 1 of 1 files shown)


amacneil:
What's the purpose? I looked and don't see any previous dependabot PRs that this would have affected.

MatthijsBurgh:
@amacneil Oh, I see we are only running dependabot for GH actions here. I thought you were running NPM updates here as well. For NPM it is useful, as it reduces the number of PR significantly.

Do you want to start using dependabot for NPM updates too?

amacneil:
I'm confused, there is no NPM in this repo

MatthijsBurgh:
@amacneil my mistake.

PR #730, opened 2022-04-13
Fixed float arrays conversion

zecastelo:
**Public API Changes**
None

**Description**
Messages with `float32` arrays (bounded and unbounded) were being silently dropped. After debugging I found that the conversion was the problem. I added the fix and the respective tests.


defunctzombie:
Please rebase on main and that should resolve the lint failure.

jtbandes:
The lint failure is now a real error: you need to re-format `message_conversion.py`

jtbandes (approved):

Issue #718, opened 2022-03-07
Rosbrige websocket tries to decode BSON message into utf-8


tsender:
**Description**

When rosbridge is in `bson_only_mode` and receives a BSON message, the on_message() function in websocket_handler.py attempts to decode the message into utf-8. Doing so results in an error of the form "UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9 in position 10: invalid continuation byte".

Solution: Change the if condition from `if isinstance(message, bytes)` to `if not self.bson_only_mode and isinstance(message, bytes)` in the on_message(self, message) function definition.

     
- Library Version: ros2
- ROS Version: foxy
- Platform / OS: Ubuntu 20.04

**Steps To Reproduce**
1. Launch rosbridge with bson_only_mode set to True
2. Send rosbridge  a BSON message

**Expected Behavior**
Rosbridge should parse the message appropriately without error.

**Actual Behavior**
Error message of the form: "UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9 in position 10: invalid continuation byte".

PR #739, opened 2022-05-09
Fixed utf-8 decode exception when in bson_only_mode

JKaniarz:
**Public API Changes**

None

**Description**

Modifed Protocol.incoming to:

- accept both bytes() and str (and removed the utf-8 decode from RosbridgeWebSocket.on_message()).
- Deinterleved the BSON handling code from the JSON handling code.
- Buffer bytes() in partial BSON messages.

This is also a steppingstone toward getting TCP working again.

https://github.com/RobotWebTools/rosbridge_suite/issues/718


JKaniarz:
Is there anything standing in the way of merging this?

jtbandes:
Is there a way we can add tests for this fix? Or is that impossible now that we have only the websocket server?

JKaniarz:
I'm not familiar with your testing framework but here's what the tests would need to be for RosbridgeWebSocket:

- Send a time message in BSON format with all zero variables and make sure that it is sent to Protocol as bytes(). The trick with this message is that these BSON bytes can successfully decode as UTF-8 and we want to make sure that it isn't converted.
- Send a >127 byte BSON message over the wire (PoseWithCovarience will work) and make sure that it is sent to Protocol as bytes() and no exception is thrown. This message can't be converted to UTF-8 and we want to make sure it isn't silently dropped.

These tests would apply to the TCP and UDP connections too if you ever bring them back.

JKaniarz:
Side conversation about Protocol's buffering. I left the buffering as-is, but it's only necessary for JSON over TCP. WebSocket and UDP will naturally provide JSON and BSON at the correct length. BSON has length bytes, so it's easy to chop up the TCP stream (which is what the old TCP handler did, I copied it into Protocol).

Would you like me to trim it down? As it stands, any future protocols can just blast bytes at Protocol and Protocol will do the slicing.

JKaniarz:
Is there any interest in this?

PR #741, opened 2022-05-11
Stability improvements for multiple clients

p0rys:
Public API Changes
None

Description
This changes will improve the stability when multiple clients are connected simultaneously.

Fix 1 (07bd9b8): Changes "has_subscribers" to return the combined count of subscriptions and new subscriptions . This will avoid unregister of subscribers (see line 284-285) if one client disconnects while all others are still waiting for the initial callback (there are still in "new_subscriptions" and not in "subscriptions").

Fix 2 (227bfbe): Removes a possible deadlock if a client is unsubscribed before it receives the initial callback. In that case "self.new_subscriptions.values()" passed as "callbacks" would be empty and trigger a second "self.lock" (line 209) while still in the lock from line 232.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -137,6 +137,8 @@ class MultiSubscriber:
 
     def unregister(self):
         self.node_handle.destroy_subscription(self.subscriber)
+        if self.new_subscriber:
+            self.node_handle.destroy_subscription(self.new_subscriber)
         with self.lock:
             self.subscriptions.clear()
 
--- a/rosbridge_library/src/rosbridge_library/internal/subscribers.py
+++ b/rosbridge_library/src/rosbridge_library/internal/subscribers.py
@@ -193,7 +193,7 @@ class MultiSubscriber:
         with self.lock:
             return len(self.subscriptions) + len(self.new_subscriptions) != 0
 
-    def callback(self, msg, callbacks=None):
+    def callback(self, msg):
         """Callback for incoming messages on the rclpy subscription.
 
         Passes the message to registered subscriber callbacks.
@@ -206,9 +206,8 @@ class MultiSubscriber:
         outgoing = OutgoingMessage(msg)
 
         # Get the callbacks to call
-        if not callbacks:
-            with self.lock:
-                callbacks = self.subscriptions.values()
+        with self.lock:
+            callbacks = self.subscriptions.values()
 
         # Pass the JSON to each of the callbacks
         for callback in callbacks:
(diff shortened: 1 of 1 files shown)


zflat:
I believe this fixes the deadlock that I observed in https://github.com/RobotWebTools/rosbridge_suite/issues/744
I can reproduce my issue on `ros2` commit 8a705de24bafe103981b0dd90f8494c991529bd3
Then I switch to `p0rys:multiple_clients_fixes` commit 1c71aee4fe250b730a6afa09335a6c00581ea0a5 and I do not see the same deadlock issue.

Thanks @p0rys for this fix!

zflat (approved):
Just a note and not related to this bug fix but the instance member `self.lock` is not very descriptive about what is being locked. It would be better named as `self.subscriptions_lock` so it is more clear that this lock is needed every time the `self.subscriptions` and `self.new_subscriptions` members are accessed. Or put another way, the locking logic is kinda complex in this class and it would be nice if it was less error prone.

Anyway these changes look good to me.

TheEppicJR:
CI for Rolling was updated to Ubuntu 22.04 and should work now

zflat:
@p0rys What do you think about changing `Lock` to `RLock` (https://docs.python.org/3/library/threading.html#threading.RLock) as an alternative way to prevent deadlocks than the changes in https://github.com/RobotWebTools/rosbridge_suite/commit/227bfbea38aaab3c9c4c00d24eff0e5b6b6a09fd

p0rys:
@zflat Seems reasonable for this problem but i'm far from a python expert. Also i would rather implement #426 then putting more effort into this fix.

Alqio:
This fixes https://github.com/RobotWebTools/rosbridge_suite/issues/758

rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -137,6 +137,8 @@ def __init__(self, topic, client_id, callback, node_handle, msg_type=None, raw=F
 
     def unregister(self):
         self.node_handle.destroy_subscription(self.subscriber)
+        if self.new_subscriber:
+            self.node_handle.destroy_subscription(self.new_subscriber)
```
achim-k, line 141: This should be done inside the `with self.lock` block as the assignment of `self.new_subscriber` is also protected by the lock

rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -229,13 +230,16 @@ def _new_sub_callback(self, msg):
...
+        # transfer new subscriptions to the "regular" ones
         with self.lock:
-            self.callback(msg, self.new_subscriptions.values())
             self.subscriptions.update(self.new_subscriptions)
             self.new_subscriptions = {}
             self.node_handle.destroy_subscription(self.new_subscriber)
             self.new_subscriber = None
 
+        # send msg to client
+        self.callback(msg)
```
achim-k, line 241: This would effectively cause existing clients to receive messages twice when a new client connects. 
I think it would be better to leave it how it was and instead use an [RLock](https://docs.python.org/3/library/threading.html#rlock-objects) to avoid the potential deadlock.

achim-k (changes requested):
This PR avoids the deadlock issue, but it causes some side effects such as existing clients receiving messages twice when a new client subscribes to the topic.

@p0rys do you have time address my comments? Otherwise I can do that quickly myself

achim-k:
Superseded by #803

PR #742, opened 2022-05-11
Fixed CBOR compression

p0rys:
**Public API Changes**
None

**Description**
tl;dr: Fixes CBOR compression for ROS2 and reenables unit tests for it. This is based on #711.

The variable "_slot_types" is not provided by rclpy and therefore CBOR could not correctly determine the data types.  Based on the work from @kedus42 in #711 it will now use "get_fields_and_field_types()" and work with the new ROS2 datatypes (e.g. octet).
Unit tests in "test_cbor_conversion.py" has been updated / extended to test the new cases with the ROS2 example messages from example_interfaces pkg. The std_msgs pkg used before is depracted since foxy.


mjforan:
@achim-k @defunctzombie can we get a review of this please? Seems like a straightforward, necessary change.

mjforan:
It works for me - I am sending custom image messages with CBOR compression on ROS 2 Humble.

p0rys:
We still use this fix in our production system. It would be nice if this could be reviewed and merged.

PR #774, opened 2022-08-03
fix memory leak on service call

### LINKED ISSUE #772: CPU & Memory usage increases over time.
**Description**

CPU & Memory usage increases over time.

- Library Version:
  installed with apt
  `ros-galactic-rosbridge-suite/focal,now 1.2.0-1focal.20220520.223734 amd64 [installed]`
- ROS Version:
  galactic
- Platform / OS:
  Ubuntu20.04

**Steps To Reproduce**

1. `ros2 run demo_nodes_cpp add_two_ints_server`
   https://github.com/ros2/demos is used.
1. `ros2 run rosbridge_server rosbridge_websocket`
2. Run the following scripts in multiple terminals: (two terminals are OK)
```python3
import roslibpy
import time
import datetime
import sys
from threading import Thread, Lock

ros = roslibpy.Ros(host='localhost', port=9090)
ros.run(timeout=None)

srv = roslibpy.Service(ros, "/add_two_ints", "example_interfaces/srv/AddTwoInts")
srvreq = roslibpy.ServiceRequest({"a": 1, "b": 2})

def srvcb(data):
    print(data)

def srv_errcb(data):
    print(f"error: {data}")

def service():
    while ros.is_connected:
        srv.call(srvreq, callback=srvcb, errback=srv_errcb, timeout=5)
        time.sleep(0.001)

th = Thread(target=service, daemon=True)
th.start()

while True:
    time.sleep(1)
```
1. wait for several hours. 4 or 5 hours.
2. In the beginning, CPU usage is about 45%, but after a few hours, it will exceed 50%~70%, in my environment.

**Expected Behavior**

CPU usage should keep up.

**Actual Behavior**

CPU usage increases slightly over time.

obi-t4:
**Public API Changes**

None

**Description**

A memory leak occurs each time a service call is invoked from rosbridge client.
This means calling service call from the client leads to resource exhaustion.
This fixes https://github.com/RobotWebTools/rosbridge_suite/issues/772

https://github.com/RobotWebTools/rosbridge_suite/issues/772


lroop:
I can confirm that I have observed this same issue with rosbridge gradually consuming a lot of CPU and memory after running for several hours, causing other ROS nodes to slow down, and this PR appears to fix the issue.

rosbridge_library/src/rosbridge_library/internal/services.py
```diff
@@ -120,9 +122,14 @@ def call_service(node_handle, service, args=None):
     # Populate the instance with the provided args
     args_to_service_request_instance(service, inst, args)
 
-    client = node_handle.create_client(service_class, service)
+    if service_class in clients:
+        client = clients[service_class]
+    else:
+        client = node_handle.create_client(service_class, service)
+        clients[service_class] = client
```
achim-k: This does not work if you have multiple services with the same service type but different names. Let's say you have two identical services under a different namespace:
- `/robot_1/do_something`
- `/robot_2/do_something`

You will only call one of these services (the one you called first), as they have the same service type.

achim-k (changes requested):
This does not work if you have multiple services with the same service type. I would propose to simply [destroy](https://docs.ros2.org/galactic/api/rclpy/api/services.html#rclpy.service.Service.destroy) the service client after the call.

rosbridge_library/src/rosbridge_library/internal/services.py
obi-t4: I'll consider it, thanks!

obi-t4: @achim-k 
> I would propose to simply [destroy](https://docs.ros2.org/galactic/api/rclpy/api/services.html#rclpy.service.Service.destroy) the service client after the call.

Calling `rclpy.service.Service.destroy` is user responsibility, so I think it cannot be used in the library.
And I know that `node_handle.destroy_client(client)` does not free all memory.

obi-t4: @achim-k https://github.com/RobotWebTools/rosbridge_suite/pull/774/commits/1909996867cc757253030734a27f7c1c1d2378b5 should work.

obi-t4:
@achim-k 
We must fix this memory leak issue for service call. Otherwise service call can't be used.

rosbridge_library/src/rosbridge_library/internal/services.py
```diff
@@ -120,9 +122,15 @@ def call_service(node_handle, service, args=None):
...
-    client = node_handle.create_client(service_class, service)
+    client_key = f"{service}:{service_type}"
+    if client_key in clients:
+        client = clients[client_key]
+    else:
+        client = node_handle.create_client(service_class, service)
+        clients[client_key] = client
 
     result = client.call(inst)
+
```
achim-k, line 126:
```suggestion
    node_handle.destroy_client(client)
```

achim-k (commented):
I think the simplest solution would be to destroy the service client again after use. Would that solution be ok with you?

rosbridge_library/src/rosbridge_library/internal/services.py
obi-t4: @achim-k

No.
> And I know that node_handle.destroy_client(client) does not free all memory.

Adding this code, memory leak issue is not solved.

obi-t4: My first commit was this, but I found it didn't release all the memory.

obi-t4: @achim-k 

releasing resources by `node_handle.destroy_client(client)` is rclpy responsibility, so I'll change this PR, and I'll also submit to fix this issue to rclpy repo.

obi-t4:
@achim-k PR has been updated, please have a check!

rosbridge_library/src/rosbridge_library/internal/services.py
achim-k: If you create a `rclpy` issue / PR, could you add a link here? Thanks!

achim-k (approved):

obi-t4:
@achim-k The isue: https://github.com/ros2/rclpy/issues/1034 has been created.

Issue #736, opened 2022-04-29
ros2 rosbridge_websocket, if use param --address,  there will be errors


JunlingQin:
**Description**

- Library Version:
- ROS Version: ros2
- Platform / OS: Ubuntu 20.04.4

**Steps To Reproduce**

Traceback (most recent call last):
  File "/opt/ros/galactic/lib/rosbridge_server/rosbridge_websocket", line 335, in <module>
    main()
  File "/opt/ros/galactic/lib/rosbridge_server/rosbridge_websocket", line 320, in main
    node = RosbridgeWebsocketNode()
  File "/opt/ros/galactic/lib/rosbridge_server/rosbridge_websocket", line 169, in __init__
    address = int(sys.argv[idx])
ValueError: invalid literal for int() with base 10: ''

**Expected Behavior**

**Actual Behavior**

achim-k:
Fixed by #780

PR #780, opened 2022-09-01
hostname parameter is no longer converted as an int value

hugoperier:
**Public API Changes**

Fix of #736  
As mentioned in the documentation of tornado.netutil, hostname as to be a string, the conversion to an integer lead any usage of this parameter to an error

**Description**

#736


achim-k (approved):
LGTM

This fixes the special case when running the `rosbridge_websocket` script directly instead of launching it via the launch file. When launching it via the launch file, the `address` is passed in as ros parameter to the script.

```
ros2 launch rosbridge_server rosbridge_websocket_launch.xml address:=0.0.0.0
```

PR #781, opened 2022-09-02
Convert to ROS Python packages

srmainwaring:
**Public API Changes**

None, this is a packaging change.

**Description**

This PR converts the `rosapi`, `rosbridge_library`, and `rosbridge_server` to ROS 2 Python packages using the structure described in [Developing a ROS 2 package](https://docs.ros.org/en/foxy/How-To-Guides/Developing-a-ROS-2-Package.html#python-packages). 

The motivation is to get a working version of the `rosbridge_suite` on macOS running ROS2 Humble using the guidelines outlined in [macOS build from source in ROS2 Galactic](https://github.com/ros2/ros2/issues/1148) and associated issues. This is a pre-requisite to use Foxglove Studio as the ROS 2 native TCP connection is not working and the data source dialog recommends using the Rosbridge connection.

One of the challenges on macOS is resolving library dependencies without disabling System Integrity Protection (SIP). This seems to work well provided that the colcon / ament build system installs packages according to guidelines, and often doesn't work, with very hard to trace and fix shared library resolution issues, when it doesn't.

I expect this is something of a niche requirement, so it may be better making this available as topic branch for macOS users with a note in the README to guide them.

**Testing**

Running on ROS2 Humble with Foxglove Studio 1.23.0 (darwin) on macOS Monterey (Intel).

ros2-humble-bridge-foxglove


defunctzombie:
@srmainwaring As a related matter to this PR - I've had great success using [robostack](https://robostack.github.io/) to install ROS (1 and 2) on my mac and use rosbridge. They are pretty open to PRs to add default packages and I had them add rosbridge_suite to their humble build: https://github.com/RoboStack/ros-humble/pull/7

Might be worth looking at for macOS workflows if you want pre-built packages.

srmainwaring:
> I've had great success using [robostack](https://robostack.github.io/) to install ROS (1 and 2) on my mac and use rosbridge

@defunctzombie good suggestion - RoboStack has come a long way with macOS support since I last looked. Worth looking into further to see if I can help get Gazebo Garden available on the ros-humble channel - that could be a huge help to my workflow.

zflat:
I think that there is an underlying problem with `ament` if packages are not installed correctly to work with macOS. Changing the module distribution configuration for these packages alone seems like a hack and does not address the underlying issue. Fixing the underlying issue would not only benefit rosbridge but any other ROS2 python package configured with ament CMake. 

I suggest closing this merge request and filing a bug with https://github.com/ament/ament_cmake/issues unless an issue already exists that describes the problem you are seeing @srmainwaring.

defunctzombie:
> I think that there is an underlying problem with `ament` if packages are not installed correctly to work with macOS.

@zflat My reading of this PR and @srmainwaring description is that he is claiming this package does not conform to the documented way of creating a python page here: https://docs.ros.org/en/humble/How-To-Guides/Developing-a-ROS-2-Package.html#python-packages leading to `ament` not being able to do its job because the package does not match what ament requires.


> I suggest closing this merge request and filing a bug with https://github.com/ament/ament_cmake/issues unless an issue already exists that describes the problem you are seeing @srmainwaring.

If `ament` has a specific way a package should be organized to install correctly, then it seems the issue is with rosbridge not conforming. It would be appropriate to file a _feature_ request asking for `ament` to support additional methods of organization.

However, if the package structure used in `rosbridge` today _is_ valid and something `ament` is _meant_ to support then I agree, this is a deficiency in `ament` and would be a bug.

PR #789, opened 2022-09-23
Remove unnecessary checking of topic globs on every message callback

achim-k:
**Public API Changes**
None

**Description**
Topic globs are currently checked on every message callback. This is however not necessary, as we check topic globs already at the point when a subscription request is made. If the topic is not allowed to be subscribed, no subscriber is created and the message callback is never called.

This PR removes unnecessary topic glob checking which results in a marginal performance improvements (especially for high-rate topics)

I think it makes sense to backport this to `ros1` branch as well


achim-k:
CI failure is unrelated. Looks like CI pipeline is broken

achim-k:
Closing this one and opening a PR against `ros1`.

Edit: #793

PR #792, opened 2022-09-28
Skip unnecessary conversion for cbor/cbor-raw compression

achim-k:
**Public API Changes**
None

**Description**
This change avoids some unnecessary conversions when using `cbor`/`cbor-raw` compression, leading to a significantly perfomance boost.
Additionally, the cbor encoded message is now cached, to avoid encoding for every client that is subscribed to the topic.


achim-k:
CI failure is unrelated, seems to fail for all builds at the moment.

achim-k:
@jtbandes could you PTAL?

jtbandes (approved):

PR #813, opened 2022-10-16
Add ros2 action client capability 

sathak93:
**Public API Changes**

 #### 3.3.10 Send Goal

```json
{ "op": "send_goal",
  (optional) "id": <string>,
  "action_name": <string>,
  "action_type": <string>,
  (optional) "feedback": <bool>,
  (optional) "goal_msg": <list<json>>
}
```

This command creates an action client with the specified action name and type if not already created; sends the goal msg to the action server and returns  the feedback and result of the goal.

 * **action_name** – the name of the action to which the goal will be sent (ex. /navigate_to_pose)
 * **action_type** – the type of the action (ex. /nav2_msgs/action/NavigateToPose)
  * **feedback** – an optional boolean specifies whether to receive the feedback for this goal or not. defaults to true)
 * **goal_msg** – if the Goal has no args, then goal msg does not have to be
    provided, though an empty list is equally acceptable. goal msg should be a list of json objects representing the arguments to the action
 * **id** – an optional id to distinguish this goal call

 #### 3.3.11 Cancel Goal

```json
{ "op": "cancel_goal",
  (optional) "id": <string>,
  "action_name": <string>
}
```

This command cancels all the executing and pending goals of the specified action and return the result of cancel call.

 * **action_name** – the name of the action (ex. /navigate_to_pose)
 * **id** – an optional id to distinguish this cancel goal call.

 #### 3.3.12 Destroy Client

```json
{ "op": "destroy_client",
  (optional) "id": <string>,
  "action_name": <string>
}
```
This command destroys the action client if the action client was created on earlier send_goal calls.

 * **action_name** – the name of the action (ex. /navigate_to_pose)
 * **id** – an optional id to distinguish this cancel goal call.

**Description**
This pr adds ros2 action client capability using the direct approach. now clients can create action clients, send goal to get result and feedback, cancel goal, destroy client.

this address #697


defunctzombie:
This PR looks like it is proposing additions to the spec. For changes to the spec please update the spec document in your PR and describe the proposed changes in the public api changes section of the PR description.

Reading #697 and seeing that actions and services are implemented on top of topics I am not clear on what capability this adds that could not be accomplished but it does increase the maintenance surface area. Please talk a bit about that so we can evaluate the tradeoff.

achim-k:
I believe that the best way to add action support is to do that on the client side. [Roslibjs](https://github.com/RobotWebTools/roslibjs) for instance provides an [ActionClient](http://robotwebtools.org/roslibjs/ActionClient.html) and a [SimpleActionServer](http://robotwebtools.org/roslibjs/SimpleActionServer.html), not sure if they work for `ros2` actions though.
Overall I would prefer to keep the rosbridge backend lean and maintanable.

nakai-omer:
@defunctzombie This is critical as rosbridge currently throws exceptions when trying to use an Action. It looks for actions as if they are msgs:
```
[rosbridge_websocket-18] 1669043713.492345222 [rosbridge_websocket] [ERROR] [Client 975e4ce5-a080-45b5-ab09-446a0470b0e5] [id: subscribe:/exploration_command/result:70] subscribe: Unable to import msg class ExplorationActionResult from package nakai_interfaces. Caused by module 'nakai_interfaces.msg' has no attribute 'ExplorationActionResult'
```

Community is also in need of this:
https://answers.ros.org/question/403018/is-there-a-workaround-for-ros2-actions-in-rosbridge/
https://answers.ros.org/question/404992/rosbridge_websocket-unable-to-import/?comment=409659#post-id-409659

With that said @achim-k, trying out your PR we are still getting the following errors:
```
[ERROR] [Client d8eb1d55-acf7-432a-af10-65d23086a715] [id: subscribe:/exploration_command/result:65] subscribe: Unable to import action class ExplorationActionResult from package nakai_interfaces. Caused by module 'nakai_interfaces.action' has no attribute 'ExplorationActionResult'
[ERROR] [Client d8eb1d55-acf7-432a-af10-65d23086a715] [id: advertise:/exploration_command/goal:66] advertise: Unable to import action class ExplorationActionGoal from package nakai_interfaces. Caused by module 'nakai_interfaces.action' has no attribute 'ExplorationActionGoal'
[ERROR] [Client d8eb1d55-acf7-432a-af10-65d23086a715] [id: subscribe:/exploration_command/feedback:69] subscribe: Unable to import action class ExplorationActionFeedback from package nakai_interfaces. Caused by module 'nakai_interfaces.action' has no attribute 'ExplorationActionFeedback'
[ERROR] [Client d8eb1d55-acf7-432a-af10-65d23086a715] [id: subscribe:/exploration_command/result:70] subscribe: Unable to import action class ExplorationActionResult from package nakai_interfaces. Caused by module 'nakai_interfaces.action' has no attribute 'ExplorationActionResult'
 ```

**Comment by sathak93** [comment]
@nakai-omer  you are using actionclient in roslibjs which only works for ros1 as it simply subscribes to feedback and result and publish goal topics. for ROS 2 action client is not added yet . you can use the code in the  folder ros2action  on  my fork https://github.com/sathak93/roslibjs/tree/ros2actionclient. once this PR is finalized and added to rosbridge_server . pr for client side implementation can be added.

**Comment by nakai-omer** [comment]
@sathak93 Got it thanks a lot!

**Comment by amalnanavati** [comment]
@sathak93 Your [rosbridge_suite code](https://github.com/sathak93/rosbridge_suite/tree/act2) and [roslibjs code](https://github.com/sathak93/roslibjs/tree/ros2actionclient) was a lifesaver! Looking forward to this being incorporated into the official releases. 🙌

**Comment by ahoenerBE** [comment]
Is this PR still active? It's been a while and seems like it'd be a good addition

**Comment by sathak93** [comment]
@ahoenerBE  and @amalnanavati  it is still in good condition i think. rosbridge doesn't have action capability in ros 2 only.  Most of the current users of rosbridge are using ROS 1 i think thats why they doesn't know they miss out important feature in ros2  .  This PR will increase code base. if it is okay with maintainers then we can review and merge
> I believe that the best way to add action support is to do that on the client side. [Roslibjs](https://github.com/RobotWebTools/roslibjs) for instance provides an [ActionClient](http://robotwebtools.org/roslibjs/ActionClient.html) and a [SimpleActionServer](http://robotwebtools.org/roslibjs/SimpleActionServer.html), not sure if they work for `ros2` actions though. Overall I would prefer to keep the rosbridge backend lean and maintanable.

**Comment by mjforan** [comment]
I need this capability as well.

@defunctzombie  Can we get another review of this?

**Comment by sea-bass** [comment]
+1 -- we are looking to add ROS 2 Action Server support soon and would love for this PR to be in first as well.

**Comment by ahoenerBE** [comment]
> Reading https://github.com/RobotWebTools/rosbridge_suite/issues/697 and seeing that actions and services are implemented on top of topics I am not clear on what capability this adds that could not be accomplished but it does increase the maintenance surface area.  

> I believe that the best way to add action support is to do that on the client side. [Roslibjs](https://github.com/RobotWebTools/roslibjs) for instance provides an [ActionClient](http://robotwebtools.org/roslibjs/ActionClient.html) and a [SimpleActionServer](http://robotwebtools.org/roslibjs/SimpleActionServer.html), not sure if they work for `ros2` actions though. Overall I would prefer to keep the rosbridge backend lean and maintanable.

I appreciate and understand the desire to keep the codebase maintainable, as too many changes can create unneeded bloat, but I don't understand the pushback for supporting a core feature of ROS2. Sure, it's publishers and service calls under the hood, but that way of thinking requires each library in different languages to write some sort of adapter for ROS2 actions themselves.  
It's creating much more bloat in 10 other projects to save a bit of code in the core one.  

And for that matter, if actions are ever changed, then each of those projects will have to update their implementation to restore functionality, instead of the base rosbridge being updated and having the spec maintain functionality for all libraries that interact with it automatically.  

I think it makes much more sense to add something that can natively interact with the ROS2 action server, and provide the *standard* for other libraries to accept, rather than making each different library do their own thing

**Comment by sea-bass** [comment]
> Most of the current users of rosbridge are using ROS 1

I think it stems from this. In ROS 1 you could interact with actions using their underlying topics, but as far as I know that is not the case in ROS 2 and we do need to use the action interface.

**Comment by amalnanavati** [comment]
Yup, ROS2 implements actions using hidden services/topics, and the service/message types are not easily introspectable ([relevant design doc](https://design.ros2.org/articles/actions.html)). Both those factors prevented me from being able to interact with ROS2 actions using topic/service calls in roslibjs, which is why I think we need functionality on the rosbridge level directly.

**Comment by okahilak** [comment]
+1 to supporting ROS2 actions in rosbridge and getting this patch forward. Not being able to use actions in the front-end is currently a major blocker in our project, and as @amalnanavati pointed out, actions are not easily doable in ROS2 using topic/service calls.

**Comment by sea-bass** [comment]
BTW I am going to work on action server functionality fairly soon (in the next few weeks), so it would be fantastic if this got merged in. I'm happy to also help write unit tests for this action client functionality as it's something we'll likely want at some point too.

File: ROSBRIDGE_PROTOCOL.md [hunk 1]
```diff
@@ -467,6 +467,51 @@ A response to a ROS service call
...
+
+This command cancels all the executing and pending goals of the specified action and return the result of cancel call.
+
+ * **action_name** – the name of the action (ex. /navigate_to_pose)
+ * **id** – an optional id to distinguish this cancel goal call.
+
+ #### 3.3.12 Destroy Client
+
+```json
+{ "op": "destroy_goal",
```
sea-bass:
```suggestion
{ "op": "destroy_client",

File: ROSBRIDGE_PROTOCOL.md [hunk 2]
```diff
@@ -467,6 +467,51 @@ A response to a ROS service call
...
+ #### 3.3.10 Send Goal
+
+```json
+{ "op": "send_goal",
+  (optional) "id": <string>,
+  "action_name": <string>,
+  "action_type": <string>,
+  (optional) "feedback": <bool>,
+  (optional) "goal_msg": <list<json>>
+```
```
sea-bass: you're missing the closing bracket `}` in all of these

rosbridge_library/src/rosbridge_library/capabilities/action_client.py
```diff
@@ -0,0 +1,184 @@
...
+    def __init__(self, protocol):
+        # Call superclass constructor
+        Capability.__init__(self, protocol)
+
+        # Register the operations that this capability provides
+        protocol.register_operation("send_goal", self.send_goal)
+        protocol.register_operation("destroy_client", self.destroy_client)
+        protocol.register_operation("cancel_goal", self.cancel_goal)
+
+        self._actionclients = {}
```
sea-bass: Consider calling this `self._action_clients`

rosbridge_library/src/rosbridge_library/internal/ros_loader.py
```diff
@@ -143,7 +167,29 @@ def _get_srv_class(typestring):
...
+    """If not loaded, loads the specified action class then returns an instance
+    of it
+
+    Throws various exceptions if loading the action class fails"""
+    global _loaded_actions, _actions_lock
+    try:
+        # The type string starts with the package and ends with the
+        # class and contains module subnames in between. For
+        # compatibility with ROS1 style types, we fall back to use a
+        # standard "srv" subname.
```
sea-bass:
```suggestion
        # standard "action" subname.

File: rosbridge_library/src/rosbridge_library/capabilities/action_client.py [hunk 5]
```diff
@@ -0,0 +1,184 @@
+import fnmatch
+from functools import partial
+
+from rosbridge_library.capability import Capability
+from rosbridge_library.internal.actions import ActionClientHandle, GoalHandle
+from rosbridge_library.internal.message_conversion import extract_values
+
+
+class ActionClientRequests(Capability):
```
sea-bass, line 9: This doesn't seem like it's named consistently as a "capability" compared to the others.

I think maybe this should be called `SendActionGoal` or something?

sea-bass (commented):
Looks good, though I didn't test.

Would be great if you also got to work on some unit tests (I just ported all the pub/sub ones to ROS 2 in https://github.com/RobotWebTools/rosbridge_suite/pull/882 if you want a reference)... but I'm also happy to help do this after this one is merged.

BTW I am only reviewing as an interested party, sadly I don't have merging power here.

defunctzombie:
@sea-bass 👋 Good seeing you at roscon. Since Foxglove doesn't leverage rosbridge anymore, our support of this package has waned. You mentioned that PickNik relies on rosbridge for some products so I am open to adding you as a maintainer in Github to let you all carry the torch forward.

sea-bass:
> @sea-bass 👋 Good seeing you at roscon. Since Foxglove doesn't leverage rosbridge anymore, our support of this package has waned. You mentioned that PickNik relies on rosbridge for some products so I am open to adding you as a maintainer in Github to let you all carry the torch forward.

This would be great, thank you for the offer -- could we also add my colleague @ezrabrooks, who is also a contributor to this repo?

ahoenerBE:
> Since Foxglove doesn't leverage rosbridge anymore, our support of this package has waned.  

@sea-bass @defunctzombie Is the recommended bridge node now the foxglove ws-protocol if support here is waning? I want to make sure our projects are using something that will be updated in the future

sea-bass:
I'd defer to Foxglove's answer, but from what I heard at ROSCon Foxglove uses a different protocol which means you can definitely use it on your end -- it is open-source -- but you'll need to modify/rewrite your own client libraries.

At least for the foreseeable future, PickNik plans to continue using this protocol (just the ROS 2 branch though), and we're happy to help maintain.

ahoenerBE:
> I'd defer to Foxglove's answer, but from what I heard at ROSCon Foxglove uses a different protocol which means you can definitely use it on your end -- it is open-source -- but you'll need to modify/rewrite your own client libraries.
> 
> At least for the foreseeable future, PickNik plans to continue using this protocol (just the ROS 2 branch though), and we're happy to help maintain.

@sea-bass That's good to know. I'm maintaining rosreact, which is a react hooks wrapper around roslibJS and it'd be unfortunate to have to rewrite it all for a new comms protocol. I'm also open to helping support, if people are wanting to keep this package alive.

amacneil:
@sea-bass @EzraBrooks I sent you both invites to this github org.

As for the ROS-Foxglove Bridge, we designed it to avoid performance and other issues we were seeing with rosbridge. It's open source, but we don't recommend it for general purpose (i.e. non-Foxglove related) use.

sea-bass:
@sathak93 I didn't want to make you write a ton of unit tests, so wanted to share that I'm taking pieces from the great work you've done here and implementing them a little different + with tests + also adding action servers in https://github.com/RobotWebTools/rosbridge_suite/pull/886

If you have any inputs please let me know.

Also, my plan is to try do the `roslibjs` changes next, and I saw you have a nice branch there. Do you mind if I try work with that as well once I get there?

Thank you!

sea-bass:
Closing in favor of https://github.com/RobotWebTools/rosbridge_suite/pull/886, which supports both action servers and clients and is now merged!

Thanks again for your reference client implementation, @sathak93 -- it was very useful to my implementation and you should 100% get credit too.

Issue #810, opened 2022-10-13
Adding namespace in the launch file is not working as expected 


hardesh-peppermint:
**Description**
Adding a namespace does not work with rosapi node. 

- Library Version: Latest
- ROS Version: Humble
- Platform / OS: Ubuntu 22

**Steps To Reproduce**
Add a namespace while launching the rosapi node using 

**Expected Behavior**
Everything including nodes, topics, services should be inside that namespace. 

**Actual Behavior**
At the moment, only the nodes and some of the services are going inside the namespace. The `rosapi` topics are services are node going inside the namespace. https://github.dev/RobotWebTools/rosbridge_suite/blob/413f844a27281bc02685b600c1756ffecd75a4e8/rosapi/scripts/rosapi_node#L101 in this file they've been written with global namespace. Not sure if directly removing the namespace will solve this issue as it might break other things. 

Sharing my modified launch file with namespace: 
[rosbridge_websocket_launch.xml.txt](https://github.com/RobotWebTools/rosbridge_suite/files/9776967/rosbridge_websocket_launch.xml.txt)

![IMG_0190](https://user-images.githubusercontent.com/85165012/195607205-c42c3a9b-0765-4184-8a45-bdd1fbf290ea.jpg)

Sharing some other screenshots:
![image](https://user-images.githubusercontent.com/85165012/195606910-40f286ce-d775-43ac-81a1-695dc96db699.png)

achim-k:
You are right, it should be possible to launch the rosapi / rosbridge_server nodes in a different namespace. The problem is however, that many websocket clients assume that the `rosapi` node is under the global namespace. For roslibjs this is the case in the following places: https://github.com/RobotWebTools/roslibjs/search?q=%2Frosapi

Feel free to submit a PR that removes the leading `/` for all services created [here](https://github.com/RobotWebTools/rosbridge_suite/blob/df385330dd60d99ff2875eb6c5ac0beda6290627/rosapi/scripts/rosapi_node#L91-L130) and a corresponding PR on the [roslibjs](https://github.com/RobotWebTools/roslibjs) repo. Even with a custom namespace, everything should still work since `rosapi` and `rosbridge_server` are launched in the same namespace.

hardesh-peppermint:
Okay, thanks. I'll open up a PR. Will this be merged into humble branch?

hardesh-peppermint:
Hi, 
I've opened a new PR. Is it necessary to open a new PR in roslibjs? 
[...]

PR #815, opened 2022-10-18
Removed global namespace from rosapi

hardesh-peppermint:
**Public API Changes**
None

**Description**
Removed the global namespace from rosapi. Main motivation is mentioned in #810 

Adding namespace in launch file was not working as expected


achim-k:
All new features should have a unit test. Could you write a unit test that tests this feature?

PR #838, opened 2023-02-16
Change rosbridge_suite links to ros1 branch

Combinacijus:
Currently rosbridge_suite links points to the default main branch which is "ros2", meanwhile readme is specifically for "ros1" so I updated links to reflect that


Combinacijus:
CI failure has nothing to do with my changes

PR #843, opened 2023-02-27
fix encode Exception a bytes-like object is required, not 'str'

deepinbubblegum:
- fix encode "a bytes-like object is required, not 'str'"


abishek:
Is it possible to get the logs again for this? The PR seems sensible. But maybe the error logs point to something that's not obvious?

PR #847, opened 2023-03-03
Register `call_service` operation to run in its own thread

### LINKED ISSUE #825: Service calls on same websocket are executed sequentially, not simultaneously
On a single websocket connection, service calls are executed in sequence. I expect service calls on the same websocket connection to run simultaneously.

I'm not 100% sure if this is a bug or a feature. I found this comment here ["Run service caller in the same thread."](https://github.com/RobotWebTools/rosbridge_suite/blob/7af3683ec6a02b1569d7e3c42367721b0be2ac78/rosbridge_library/src/rosbridge_library/capabilities/call_service.py#L99) which makes it seem on purpose.

I suspect this causes the IncomingQueue to block. We could maybe [change `run()` to `start()` ](https://github.com/RobotWebTools/rosbridge_suite/blob/7af3683ec6a02b1569d7e3c42367721b0be2ac78/rosbridge_library/src/rosbridge_library/capabilities/call_service.py#L100) here, but I'm not sure if this causes other things to break.

- Library Version: 1.3.1-1jammy.20221024.133759
- ROS Version: humble
- Platform / OS: Ubuntu 22.04

sea-bass:
**Public API Changes**
None

**Description**
I have changed the operation registered for `call_service` so that it's not blocking -- that is, the `CallService.call_service()` implementation now runs in its own separate thread.

This enables us to process multiple service calls in parallel, whereas right now the tools do them sequentially.

If folks are apprehensive about this blanket change, I could also envision allowing both options, so we could have a `call_service_blocking` operation that works as before, and a `call_service` that uses the changes from this PR... or some other way to configure things. Open to feedback!

Fixes #825

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
@@ -53,10 +53,19 @@ class CallService(Capability):
         Capability.__init__(self, protocol)
 
         # Register the operations that this capability provides
-        # This calls the service in a separate thread so multiple services can be processed simultaneously.
-        protocol.register_operation(
-            "call_service", lambda msg: Thread(target=self.call_service, args=(msg,)).start()
-        )
+        call_services_in_new_thread = protocol.node_handle.get_parameter(
+            "call_services_in_new_thread"
+        ).get_parameter_value()
+        if call_services_in_new_thread:
+            # Calls the service in a separate thread so multiple services can be processed simultaneously.
+            protocol.node_handle.get_logger().info("Calling services in new thread")
+            protocol.register_operation(
+                "call_service", lambda msg: Thread(target=self.call_service, args=(msg,)).start()
+            )
+        else:
+            # Calls the service in this thread, so services block and must be processed sequentially.
+            protocol.node_handle.get_logger().info("Calling services in existing thread")
+            protocol.register_operation("call_service", self.call_service)
 
     def call_service(self, message):
         # Pull out the ID
--- a/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
@@ -32,6 +32,7 @@
 
 import fnmatch
 from functools import partial
+from threading import Thread
 
 from rosbridge_library.capability import Capability
 from rosbridge_library.internal.services import ServiceCaller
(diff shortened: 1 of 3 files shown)


sea-bass:
@achim-k what do you think of this change? 

I'm particularly interested in: Should it go in as-is, or should we have some configurability on whether to have service calls block vs. run in their own threads?

rosbridge_library/src/rosbridge_library/capabilities/call_service.py
```diff
@@ -52,7 +53,10 @@ def __init__(self, protocol):
         Capability.__init__(self, protocol)
 
         # Register the operations that this capability provides
-        protocol.register_operation("call_service", self.call_service)
+        # This calls the service in a separate thread so multiple services can be processed simultaneously.
+        protocol.register_operation(
+            "call_service", lambda msg: Thread(target=self.call_service, args=(msg,)).start()
```
achim-k:
```suggestion
            "call_service", lambda msg: Thread(target=self.call_service, args=[msg]).start()
```
Not sure if it's valid, but I find this more readable

achim-k (commented):
> @achim-k what do you think of this change?
> 
> I'm particularly interested in: Should it go in as-is, or should we have some configurability on whether to have service calls block vs. run in their own threads?

Looks like a good improvement and changes are minimal :+1: 

Do you expect that change to break any existing use case?
I think it would make sense to have some sort of configurability (maybe a parameter) so users can disable the new behavior if, for whatever reason, this causes problems.

sea-bass:
> Do you expect that change to break any existing use case?

I thought about this and there are two things that come to mind.

1. If the `call_service` implementation actually returned something... which it doesn't. So we're safe here.
2. If there is any kind of "post-execution" callback where some code is run at the end of the `call_service` operation, then yes, it might break things. Because this operation now ends when the thread is started, and not when the service actually returns a response.

Anyways, I will work on making this configurable. Should this be an extra arg in the `call_service_msg_fields`, or simply a separate operation to register? Like `call_service` vs. `call_service_thread`?

achim-k:
> Should this be an extra arg in the `call_service_msg_fields`, or simply a separate operation to register? Like `call_service` vs. `call_service_thread`?

Both options would require making changes to the [protocol specification](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md) and then adding support for this to the various client libraries (roslibjs, ...). How about just making this a server configuration for now?

E.g.
```py
if call_services_in_thread:
  protocol.register_operation("call_service", lambda msg: Thread(target=self.call_service, ...
else:
  protocol.register_operation("call_service", self.call_service)
```

sea-bass:
@achim-k What do you think of this update? Added a `call_services_in_new_thread` node parameter that can be configured at launch time. Defaults to False so it's not breaking.

achim-k (approved):

achim-k:
> @achim-k What do you think of this update? Added a `call_services_in_new_thread` node parameter that can be configured at launch time. Defaults to False so it's not breaking.

Looks good, nice improvement

PR #854, opened 2023-04-05
Modifies the use of __slots__ for get_fields_and_field_types method

Voldivh:
**Public API Changes**

None

**Description**

The attribute `__slots__` from a message represents all the attributes from the python class for that message, not only the field names of each component from the message structure. So far, this shouldn't impose any issue, however, taking into account the upcoming modifications from [this PR](https://github.com/ros2/rosidl_python/pull/194) in the `rosidl_python` package, the current use of the attribute in the `rosapi/objectutils.py` file would be affected and therefore, break something. 

This PR solves that issue by modifying the use of `__slots__` for the appropriate method to retrieve the field names and types `get_fields_and_field_types()`.


achim-k:
@Voldivh is this PR still relevant? Tests currently do not fail for rolling / iron but maybe none of the existing tests covers that change. Would you mind rebasing such that all CI tests are run again?

PR #857, opened 2023-04-28
Interpret call_services_in_new_thread as boolean

rwhitney456:
**Public API Changes**

None

**Description**

I ran into an issue with having the `call_service` operation run in its own thread  (introduced in #847).  This is not a huge deal as there is a configuration option to turn it off.

However, changing the value of `call_services_in_new_thread` parameter in the `rosbridge_websocket_launch.xml` has no effect on whether or not that functionality is used.

The issue is the parameter object always evaluates to `True` instead of evaluating the value of the parameter object (which can be interpreted as a string, double, bool, etc.).  This change explicitly specifies how it should be interpreted which is probably what was intended when the parameter was introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,4 +59,4 @@ Releasing requires push access to [RobotWebTools/rosbridge_suite](https://github
     - `bloom-release --rosdistro galactic --track galactic rosbridge_suite`
     - `bloom-release --rosdistro rolling --track rolling rosbridge_suite`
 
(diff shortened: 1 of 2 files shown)


rwhitney456:
Sorry about all the commits for re-running linting.  I didn't see a way to re-run them manually.

achim-k (approved):
Thanks for the fix

Rebased it so all CI builds are run

PR #869, opened 2023-09-13
bugfix: topic list crashes if no topics are available

SubaruArai:
**Public API Changes**
None

**Description**
This is due to the [function](https://github.com/RobotWebTools/rosbridge_suite/blob/b0d0d711931d8ab7a5f079f948ced8875b4dc5d9/rosapi/scripts/rosapi_node#L72) expecting an object unpackable to 2 iterables, but an empty iterable is returned if no topics are available.

This happens only when using glob, which might have been the reason why this bug was not caught earlier.

### DIFF
--- a/rosapi/src/rosapi/proxy.py
+++ b/rosapi/src/rosapi/proxy.py
@@ -65,9 +65,13 @@ def get_topics_and_types(topics_glob):
 
         # Add topics with unknown type messages.
         unknown_type = topics.difference([x for x, _ in topic_types])
-        return zip(* topic_types + [[x,''] for x in unknown_type])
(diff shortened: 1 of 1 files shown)


SubaruArai:
@achim-k Sorry to ping you, can you review this pr (and others too)?
Thanks in advance!

achim-k (approved):
Thank you for the patch, I confirmed that it fixes the problem

As a side note, you might be interested in [foxglove_bridge](https://github.com/foxglove/ros-foxglove-bridge) as a more maintained alternative to rosbridge

SubaruArai:
@achim-k Will definitely check, thanks!

PR #886, opened 2023-10-30
Support actions in rosbridge protocol

### LINKED ISSUE #697: Add ROS2 Actions to rosbridge protocol

sea-bass:
**Description**
This PR adds ROS 2 action support to the rosbridge protocol, using the actual `rclpy.action` module instead of relying on hidden topics/services.

**Public API Changes**
Nothing changes the existing API, but instead adds a collection of new capabilities based on action servers/clients. See the corresponding doc changes here.

Also adds a `send_action_goals_in_new_thread` parameter to match what was done for services.

Note in developing this I found https://github.com/ros2/rclpy/issues/1195, which is stopping some of the actions tests from being runnable right now.

Closes https://github.com/RobotWebTools/rosbridge_suite/issues/697


EzraBrooks (approved):
The code seems sensical to me and you obviously did your legwork w/r/t testing here. I say we merge so we can start work on the client side and fix any issues found in smaller, more targeted PRs.

PR #894, opened 2023-12-07
fix type object is not iterable error

gulerburak:
**Public API Changes**

**Description**

To fix runtime error on isaac ros mqtt ros2 bridge, added this line to the affected files.
`from __future__ import annotations`


sea-bass:
What Python version are you on that doesn't support type annotations? I think if it's below Python 3.8, I'd say we shouldn't support this on the main branch.

Otherwise, we can at least put this import under a version checking guard, like:

```python
https://stackoverflow.com/a/11887885
```

In other words, if the version is less than X, only then import `annotations` from `__future__`.

And also, seems like lint is failing. You can do `pre-commit run -a` to resolve these issues.

PR #896, opened 2023-12-11
Fix issues when canceling and unadvertising actions

sea-bass:
I found a segfault in Humble described in https://github.com/ros2/rclcpp/issues/2163#issuecomment-1850925883 with destroying action servers while an executor is processing something.

For now, this simple change resolves the issue.

---

I also added a cancellation callback to the ROS 2 action servers that get created on advertise, which properly aborts the future that is created when first executing the goal.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/advertise_action.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/advertise_action.py
@@ -35,6 +35,7 @@ from typing import Any
 
 import rclpy
 from rclpy.action import ActionServer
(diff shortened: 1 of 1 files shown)


EzraBrooks (approved):
Not destroying the action server feels sketchy, but hopefully we'll circle back to this once the bug in RCL is fixed.

sea-bass:
> Not destroying the action server feels sketchy, but hopefully we'll circle back to this once the bug in RCL is fixed.

Yeah, this is why I kept a comment with the segfault when normally I would have just deleted that line.

PR #900, opened 2024-01-12
Use ROS 2 Node Clock in Message Conversion

pac48:
**Public API Changes**
The public method is  `populate_instance` from `message_converstions.py` since it now accepts a clock as an argument. A default clock argument is used in case one is not passed in. 

**Description**
Using the ROS 2 clock from a node is beneficial when using sim time. Currently, the sim time parameter has no effect. By using the node's clock, the correct clock is automatically handled.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -269,7 +269,7 @@ def _from_object_inst(inst, rostype):
     return msg
 
 
-def _to_inst(msg, rostype, roottype, clock, inst=None, stack=[]):
+def _to_inst(msg, rostype, roottype, clock=ROSClock(), inst=None, stack=[]):
     # Check if it's uint8[], and if it's a string, try to b64decode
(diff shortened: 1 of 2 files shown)


sea-bass (dismissed):
LGTM -- thanks for these fixes!

sea-bass:
Ooh, actually the `test_message_conversion.py` file has a bunch of failures, probably just missing an argument based on your changes. Want to look at that, @pac48 ?

https://github.com/RobotWebTools/rosbridge_suite/actions/runs/7504495395/job/20431670512?pr=900#step:4:6459

pac48:
> Ooh, actually the `test_message_conversion.py` file has a bunch of failures, probably just missing an argument based on your changes. Want to look at that, @pac48 ?
> 
> https://github.com/RobotWebTools/rosbridge_suite/actions/runs/7504495395/job/20431670512?pr=900#step:4:6459

Good catch. I have added a default clock argument to the internal method as well to fix that test failure.

sea-bass (approved):

Issue #926, opened 2024-05-09
get_parameter_value not present in Jazzy for rosapi_node


MoffKalast:
**Description**
Apparently the API has changed in the new release:

```bash
Traceback (most recent call last):
  File "/opt/ros/jazzy/lib/rosapi/rosapi_node", line 41, in <module>
    from rosapi import glob_helper, objectutils, params, proxy
  File "/opt/ros/jazzy/lib/python3.12/site-packages/rosapi/params.py", line 41, in <module>
    from ros2param.api import call_get_parameters, call_set_parameters, get_parameter_value
ImportError: cannot import name 'get_parameter_value' from 'ros2param.api' (/opt/ros/jazzy/lib/python3.12/site-packages/ros2param/api/__init__.py)
[ros2run]: Process exited with failure 1

```

- Library Version: ros-jazzy-rosbridge-suite:arm64/noble 1.3.2-3noble.20240502.173308 uptodate
- ROS Version: Jazzy
- Platform / OS: Ubuntu 24.04 Server, aarch64

**Steps To Reproduce**
`ros2 run rosapi rosapi_node`

**Expected Behavior**
Rosapi doesn't crash.

**Actual Behavior**
Rosapi crashes. :smiley:

earwole1:
@MoffKalast 

We are attempting to upgrade from humble to jazzy today and this also breaks `rosbridge_server`, as in:
`ros2 launch rosbridge_server rosbridge_websocket_launch.xml` 

Looks like this change was made back in Nov 2023: see [here](https://github.com/ros2/ros2cli/pull/866)

The solution appears to be to import from rclpy:
`from  rclpy.parameter import get_parameter_value`

Tested this locally and the behavior has been resolved. Just need to add a PR

rubenanapu:
> The solution appears to be to import from rclpy:
>  `from rclpy.parameter import get_parameter_value`

Awesome.

The file I added the suggested import was:

`/opt/ros/jazzy/lib/python3.12/site-packages/ros2param/api/__init__.py`

It worked like a charm.

sea-bass:
Closed by #932 -- thanks again @Oberacda!

PR #932, opened 2024-07-08
Fix invalid import of get_parameter_value in rosapi for ROS2 Jazzy.

Oberacda:
**Public API Changes**

None

**Description**

The import ros2param.get_parameter_value is no longer valid in ROS2 jazzy.
The code was removed with PR https://github.com/ros2/ros2cli/pull/866, with the suggested alternative being the import from rclpy.
As get_parameter_value was implemented in rclpy for ROS2 versions before Jazzy, this should not break older ROS2 distros.

https://github.com/RobotWebTools/rosbridge_suite/issues/926

### DIFF
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -37,8 +37,8 @@ from json import dumps, loads
 import rclpy
 from rcl_interfaces.msg import Parameter, ParameterType, ParameterValue
 from rcl_interfaces.srv import ListParameters
-from ros2node.api import get_absolute_node_name
 from rclpy.parameter import get_parameter_value
+from ros2node.api import get_absolute_node_name
 from ros2param.api import call_get_parameters, call_set_parameters
 from rosapi.proxy import get_nodes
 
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
LGTM. This works for rolling (i checked, Iron and later) but we'll have to make another branch for humble since it does not work there.

faev999:
I am still getting that same error: " ImportError: cannot import name 'get_parameter_value' from 'ros2param.api'"
when trying to launch the rosbridge `ros2 launch rosbridge_server rosbridge_websocket_launch.xml`. I am using the bridge inside a Docker image that I build like this:
````
FROM ros:rolling-ros-core

RUN apt-get update && apt-get install -y --no-install-recommends \
    ros-${ROS_DISTRO}-rosbridge-server \
 ...

**Comment by sea-bass** [comment]
This was only merged to the repo, not released as binaries. You should clone this repo and build it from source.

**Comment by Rayman** [comment]
Can we get a new release for jazzy that includes this fix?

**Comment by faev999** [comment]
> This was only merged to the repo, not released as binaries. You should clone this repo and build it from source.

When trying to build from source, I got this error:

```
1.614 E: Unable to locate package ros-rolling-example-interfaces
1.614 ERROR: the following rosdeps failed to install
1.614   apt: command [apt-get install -y ros-rolling-example-interfaces] failed
1.616 executing command [apt-get install -y ros-rolling-example-interfaces]
```

which is being trigger by this command: `rosdep install --from-paths src --ignore-src --rosdistro=$ROS_DISTRO -y `

If I switch from Rolling to Jazzy, building from source works, and launching the bridge also works.

PR #933, opened 2024-07-08
Remove `ament_cmake_core` and `ament_cmake_ros` dependencies in message packages

sea-bass:
**Public API Changes**
None

**Description**
Rolling CI has had a warning about `ament_cmake_core` and `ament_cmake_ros` not being included in the message definition packages, which... isn't needed.


sea-bass:
Actually, never mind. This is due to https://github.com/ros-tooling/action-ros-ci/issues/844, so we can just merge through these Rolling failures for now.

PR #934, opened 2024-07-08
Remove humble from, and add Jazzy to, CI on `ros2` branch

sea-bass:
**Public API Changes**
None

**Description**
Due to the changes needed in https://github.com/RobotWebTools/rosbridge_suite/pull/932, there is a divergence between Humble and later versions.

This PR removes CI for humble on the `ros2` branch, but it will be kept on in the `humble` branch. It also adds Jazzy support and bumps the `setup-ros` action to fix CI on Ubuntu 24.04.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,10 @@ jobs:
           # NOTE: Humble does not work on the `ros2` branch, so it is tested in its own branch.
           - ros: iron
             os: ubuntu-22.04
+          - ros: jazzy
+            os: ubuntu-24.04
           - ros: rolling
-            os: ubuntu-22.04
+            os: ubuntu-24.04
 
     name: ROS 2 ${{ matrix.ros }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
We need to update the branch protection to remove the required humble check.

MatthijsBurgh:
Why doesn't Iron complain about the missing ament_cmake_ros cmake? While it is correctly defined as a dependency.

sea-bass:
> Why doesn't Iron complain about the missing ament_cmake_ros cmake? While it is correctly defined as a dependency.

There is an issue with the action we are using: https://github.com/ros-tooling/action-ros-ci/issues/844

MatthijsBurgh (approved):

MatthijsBurgh:
Why not also test Jazzy?

sea-bass:
> Why not also test Jazzy?

Good point. Let me add it.

christophebedard:
You might want to try bumping `ros-tooling/setup-ros` to `0.7.7`/`v0.7`

sea-bass:
> You might want to try bumping `ros-tooling/setup-ros` to `0.7.7`/`v0.7`

This is wonderful -- that fixed it. Thank you so much for your help @christophebedard!

Issue #947, opened 2024-08-07
WebSocketClosedError flagged periodically forever


bmartin427:
Under some conditions, a stale subscription can outlive its websocket client connection, and then continue trying to write data to that closed websocket forever, resulting in continuous console spam.

- Library Version: 1.3.2
- ROS Version: iron
- Platform / OS: Ubuntu 22.04

**Steps To Reproduce**
* Run a ROS system with periodic activity on one or more topics
* Run a rosbridge server
* Run a websocket client which subscribes to one or more of the active topics, *more than once* per topic.
* Kill and restart the websocket client multiple times in quick succession.

**Expected Behavior**
Reconnections happen successfully without complaint

**Actual Behavior**
Console warnings are printed from rosbridge at a regular interval, indefinitely:
```
WebSocketClosedError: Tried to write to a closed websocket
```
Even closing the running websocket client does not silence the complaints.

PR #948, opened 2024-08-07
Avoid stale subscription when unsubscribing during resubscription (#947)

bmartin427:
**Public API Changes**
None

**Description**
If a client asks to subscribe to the same topic more than once, then disconnects, all subscriptions from that client should be torn down, not just the newest one.  Addresses #947 .


sea-bass (approved):
Thank you!

PR #955, opened 2024-10-03
Changed type hints in ros_loader.py to use imports from Typing (backport #938)

sea-bass:
Backports #938 to the `humble` branch.


sea-bass:
I missed tagging people in one of these backports.

@MatthijsBurgh since you just went through these, one more!

MatthijsBurgh (approved):

PR #953, opened 2024-09-29
Fix action cancellation by passing status from JSON

sea-bass:
**Public API Changes**
This slightly changes the protocol, specifically that the `action_result` message now contains an additional `status` field which contains an integer corresponding to the enum in https://docs.ros2.org/latest/api/action_msgs/msg/GoalStatus.html.

This does mean that the client libraries (namely, `roslibjs` which has the only working implementation) need to be updated as well.

**Description**
As described above, this PR adds a new `status` field to the `action_result` message to handle action cancellation properly.

---

Closes #920 ... maybe?

PR #962, opened 2024-10-03
Fix action cancellation by passing status from JSON (backport #953)

sea-bass:
Backports #953 to the `humble` branch.

### DIFF
--- a/ROSBRIDGE_PROTOCOL.md
+++ b/ROSBRIDGE_PROTOCOL.md
@@ -560,15 +560,17 @@ A result for a ROS action.
   "id": <string>,
   "action": <string>,
   "values": <json>,
+  "status": <int>,
   "result": <boolean>
 }
 ```
 
  * **action** – the name of the action that was executed
+ * **id** – if an ID was provided to the action goal, then the action result will contain the ID
  * **values** – the result values. If the service had no return values, then
     this field can be omitted (and will be by the rosbridge server)
- * **id** – if an ID was provided to the action goal, then the action result will contain the ID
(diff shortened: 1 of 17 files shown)


sea-bass:
I think the failing tests here may require some of the other backports to be in first.

MatthijsBurgh (approved):
Approval based on the diff being the same as the original PR. I only approve merging this PR, when the CI has been fixed (by merging other backports first).

sea-bass:
> Approval based on the diff being the same as the original PR. I only approve merging this PR, when the CI has been fixed (by merging other backports first).

The other backports did not fix the failing test, but also

* The test has nothing do with this diff
* I built this branch locally on Rolling and it passes

I'm not sure why, but I've gone ahead and disabled the test. Let me know if this is okay.

MatthijsBurgh:
It looks like we have flaky tests

sea-bass:
> It looks like we have flaky tests

I'd call this something different than flaky if they consistently pass on Iron/Jazzy/Rolling, but seem to fail on Humble. But it's an issue nonetheless.

The only recent change that touched these files is https://github.com/RobotWebTools/rosbridge_suite/pull/958 (which was part of these backports), but it's also strange that the failures would then happen on this (unrelated) PR instead of the backport itself.

Let me try revert this change and see what CI thinks.

EDIT: Nope, reverting that PR did not help.

MatthijsBurgh (approved):

sea-bass:
Mysteriously, I pulled down a ROS 2 Humble Docker container and tried to run the failing tests I skipped in there... and they passed.

PR #963, opened 2024-10-06
fix: add url_path parameter in rosbridge_websocket_launch.xml

seanpai96:
**Public API Changes**

Added `url_path` parameter in rosbridge_websocket_launch.xml.

**Description**

The ‵url_path‵ parameter is missing in the XML file of `ros2 launch`, though it can be used in `ros2 run`.

Use case:

> `ros2 launch rosbridge_server rosbridge_websocket_launch.xml url_path:=/ws`

The client should connect to `ws://<ip>/ws` without problem.


sea-bass (approved):
LGTM -- thank you!

PR #977, opened 2024-11-27
Add support for namespaces and respawn of crashed nodes

FlorianLebecque:
Added parameters in the launch file to set a namespace and node respawn.

```bash
ros2 launch rosbridge_server rosbridge_websocket_launch.xml namespace:="ROBOT" respawn:=true
```

This allow multiple robot to host their own bridge on the same network.
The respawn allows the Node to restart if it crashes. We had some issue with the ROSAPI node crashing when trying to access "dead" node information or node with empty names.

### DIFF
--- a/rosbridge_server/launch/rosbridge_websocket_launch.xml
+++ b/rosbridge_server/launch/rosbridge_websocket_launch.xml
@@ -6,6 +6,8 @@
   <arg name="certfile" default=""/>
   <arg name="keyfile" default="" />
 
+  <arg name="namespace" default="" />
+
   <arg name="retry_startup_delay" default="5.0" />
 
   <arg name="fragment_timeout" default="600" />
@@ -22,10 +24,12 @@
   <arg name="params_glob" default="" />
   <arg name="bson_only_mode" default="false" />
 
+  <arg name="respawn" default="false"/>
+
   <arg unless="$(var bson_only_mode)" name="binary_encoder" default="default"/>
 
   <group if="$(var ssl)">
-    <node name="rosbridge_websocket" pkg="rosbridge_server" exec="rosbridge_websocket" output="screen">
+    <node name="rosbridge_websocket" namespace="$(var namespace)" pkg="rosbridge_server" exec="rosbridge_websocket" output="screen" respawn="$(var respawn)" >
       <param name="certfile" value="$(var certfile)" />
       <param name="keyfile" value="$(var keyfile)" />
       <param name="port" value="$(var port)"/>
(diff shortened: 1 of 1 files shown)


bjsowa (changes requested):
The indentation after line 30 is broken. Please revert it to the current state.

bjsowa (changes requested):
You changed the indentation from 2 spaces to 4. The only lines changed in this PR should be related to your feature

FlorianLebecque:
Yes, indeed; should be fixed now

rosbridge_server/launch/rosbridge_websocket_launch.xml
```diff
@@ -22,10 +24,14 @@
   <arg name="params_glob" default="" />
   <arg name="bson_only_mode" default="false" />
 
+  <arg name="respawn" default="false"/>
+
+  <arg unless="$(var bson_only_mode)" name="binary_encoder" default="default"/>
+
   <arg unless="$(var bson_only_mode)" name="binary_encoder" default="default"/>
```
bjsowa: This line is doubled now. @sea-bass Is this even needed? This argument does not seem to be used anywhere

bjsowa (changes requested):

rosbridge_server/launch/rosbridge_websocket_launch.xml
FlorianLebecque: I don't understand how i haven't see this.
The argument seems to be used in the rosbridge_websocket node

bjsowa: The `bson_only_mode` is used but the `binary_encoder` defined here doesn't seem to be used

bjsowa:
My proposed alternative for namespaces is to use:
```
  <push_ros_namespace namespace="$(var namespace)" />
```
before the nodes are defined.

rosbridge_server/launch/rosbridge_websocket_launch.xml
```diff
@@ -6,6 +6,8 @@
   <arg name="certfile" default=""/>
   <arg name="keyfile" default="" />
 
+  <arg name="namespace" default="ws" />
```
bjsowa: This should be empty by default to not change current behavior
```suggestion
  <arg name="namespace" default="" />
```

bjsowa (changes requested):

sea-bass (approved):
Changes as they are look good and non disruptive.

Seems like everything was already resolved with @bjsowa 's review.

bjsowa (approved):
Seems fine to me. The only issue is that rosapi node [uses global namespace for all its services](https://github.com/RobotWebTools/rosbridge_suite/blob/1de68964d2cb0687284760ba45979ec2f584a5b9/rosapi/scripts/rosapi_node#L91-L130), so spawning multiple rosapi nodes in different namespaces creates a collision in service names.
roslibjs has a similar issue where it uses global namespace for service clients:
https://github.com/search?q=repo%3ARobotWebTools%2Froslibjs%20rosapi&type=code

Unless fixed in both projects, it won't work properly.

PR #978, opened 2024-12-05
fix(params): prevent parameter retrieval crashes in ROS2

### LINKED ISSUE #972: ROS2 humble getParams
**Description**
Calling the getParams service, hangs and crashes rosapi node,

- Library Version: ros-humble-rosbridge-server/jammy,now 2.0.0-1jammy.20241008.092533 amd64
- ROS Version: humble
- Platform / OS: Ubuntu 22.04

**Steps To Reproduce**

ros2 launch rosbridge_server rosbridge_websocket_launch.xml
ros2 service call /rosapi/get_param_names rosapi_msgs/srv/GetParamNames {}

**Expected Behavior**
A msg return with all available params

**Actual Behavior**
the rosapi crashes and the calling terminal hangs
```
[rosbridge_websocket-1] [INFO] [1730726532.349308897] [rosbridge_websocket]: Rosbridge WebSocket server started on port 9090
[rosapi_node-2] Traceback (most recent call last):
[rosapi_node-2]   File "/opt/ros/humble/lib/rosapi/rosapi_node", line 329, in <module>
[rosapi_node-2]     main()
[rosapi_node-2]   File "/opt/ros/humble/lib/rosapi/rosapi_node", line 321, in main
[rosapi_node-2]     rclpy.spin(node)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/__init__.py", line 222, in spin
[rosapi_node-2]     executor.spin_once()
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 739, in spin_once
[rosapi_node-2]     self._spin_once_impl(timeout_sec)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 736, in _spin_once_impl
[rosapi_node-2]     raise handler.exception()
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/task.py", line 239, in __call__
[rosapi_node-2]     self._handler.send(None)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 437, in handler
[rosapi_node-2]     await call_coroutine(entity, arg)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 391, in _execute_service
[rosapi_node-2]     response = await await_or_execute(srv.callback, request, srv.srv_type.Response())
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rclpy/executors.py", line 107, in await_or_execute
[rosapi_node-2]     return callback(*args)
[rosapi_node-2]   File "/opt/ros/humble/lib/rosapi/rosapi_node", line 281, in get_param_names
[rosapi_node-2]     response.names = params.get_param_names(self.globs.params)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rosapi/params.py", line 201, in get_param_names
[rosapi_node-2]     params.extend(get_node_param_names(node, params_glob))
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rosapi/params.py", line 221, in get_node_param_names
[rosapi_node-2]     return _get_param_names(node_name)
[rosapi_node-2]   File "/opt/ros/humble/local/lib/python3.10/dist-packages/rosapi/params.py", line 235, in _get_param_names
[rosapi_node-2]     raise RuntimeError("Wait for list_parameters service timed out")
[rosapi_node-2] RuntimeError: Wait for list_parameters service timed out
[ERROR] [rosapi_node-2]: process has died [pid 538896, exit code 1, cmd '/opt/ros/humble/lib/rosapi/rosapi_node --ros-args -r __node:=rosapi --params-file /tmp/launch_params_ci7xodun --params-file /tmp/launch_params_677akuqu --params-file /tmp/launch_params_w_m4x5yb'].
```

Mehsias:
**Public API Changes**
None

**Description**
- Add check for node's fully qualified name to prevent deadlocks
- Add 2.0s timeout to spin_until_future_complete to avoid endless loop
- Fix crashes when retrieving parameters via Roslibjs

Fixes #972


rosapi/src/rosapi/params.py
```diff
@@ -238,7 +238,7 @@ def _get_param_names(node_name):
     request = ListParameters.Request()
     future = client.call_async(request)
     if _node.executor:
-        _node.executor.spin_until_future_complete(future)
+        _node.executor.spin_until_future_complete(future, timeout_sec=2.0)
```
sea-bass: Changes look good, but this 2.0 may not hold in all cases. Wonder if should at least be a named constant instead of hard-coded, or even better an input argument to this function that comes from a configurable ROS parameter of the node?

sea-bass: Also, would it help to add this timeout in the `else` case right below this as well, which does the same but with the global executor?

Mehsias: @sea-bass That's a good idea! I've also noticed that in params.py there is a todo saying to remove the node entirely and use rosapi instead. What do you think this would entail in terms of work?

Sorry I am fairly new to ROS and don't have much experience in its uses outside of my limited scope.

sea-bass: I think the comment is instead of creating a new ROS node, one could use the existing ROS node and prevent deadlocks by using a `MultiThreadedExecutor` (which many people try to avoid as it's buggy/not performant), or make better use of the async capabilities so calling a parameter retrieval service is non-blocking.

I think it would be some work, but nothing too extreme. But certainly can be done in a follow-on PR.

bjsowa: I made a small proof of concept [here](https://github.com/RobotWebTools/rosbridge_suite/commit/65eb74413060aa3068f132bb22d4a2e279de8644) where I removed the `rosapi_params` node and used async/await features in rclpy for listing parameter names while still using `SingleThreadedExecutor`

Mehsias: @bjsowa looks good! seems compatible to what I tried to do with the last push. Although I can't test it atm because my vscode always imports from /opt/.. instead of the local scripts in the workspace.

Mehsias:
@sea-bass I don't know what I did to break the whole CI. I force pushed a squashed commit because the linter check was failing, and now everything is failing.

bjsowa:
You somehow broke the `rosbridge_websocket` symlink. I can fix it if you want.

rosapi/src/rosapi/params.py
```diff
@@ -226,21 +234,21 @@ def _get_param_names(node_name):
...
-    if node_name == _parent_node_name:
+    if node_name == _parent_node_name or node_name == _node.get_fully_qualified_name():
         return []
 
     client = _node.create_client(ListParameters, f"{node_name}/list_parameters")
 
-    ready = client.wait_for_service(timeout_sec=5.0)
+    ready = client.wait_for_service(timeout_sec=_timeout_sec)
     if not ready:
         raise RuntimeError("Wait for list_parameters service timed out")
```
bjsowa: This function can crash if the `wait_for_service` call fails as the RuntimeError is not catched anywhere. The easiest way to make this call fail is to have a node which doesn't start it's parameter services. Here's an example:
```Python
import rclpy
rclpy.init();
node = rclpy.create_node('rosapi_test_node', start_parameter_services=False)
rclpy.spin(node)
```
As this is still related to the issue you are trying to fix (crashes during parameter retrieval), could you also fix that?

IMO we shouldn't wait 5 seconds for service to become available as this will greatly increase the response time when we are running nodes without parameter servers. If we assume the rosapi node has enough time to spin and gather information about the ROS graph, we should be good with just returning an empty list if the `list_parameters` service is not available right away:
```suggestion
    if not client.service_is_ready():
        return []
```

bjsowa (changes requested):

rosapi/src/rosapi/params.py
bjsowa: Hope you don't mind me changing this myself

bjsowa (approved):

rosapi/src/rosapi/params.py
Mehsias: hey @bjsowa! Thanks so much for finishing this up. I've had the flu and wasn't able to get anything done.

PR #984, opened 2024-12-22
Add timeout option to call_service messages

bjsowa:
**Public API Changes**
Adds `timeout` option to `call_service` messages

**Description**
If during calling a service, the Service Server dies, the ServiceCaller will get stuck waiting for response. This also makes the whole protocol stuck if `call_services_in_new_threads` option is not set.

This PR adds a response timeout parameter to ServiceCaller and adds an optional `timeout` option to `call_service` messages with a default value of 5 seconds.

If the API change gets accepted I will finish this PR by adding tests that cover this feature

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
@@ -81,6 +81,7 @@ class CallService(Capability):
         fragment_size = message.get("fragment_size", None)
         compression = message.get("compression", "none")
         args = message.get("args", [])
+        timeout = message.get("timeout", 5.0)
 
         if CallService.services_glob is not None and CallService.services_glob:
             self.protocol.log(
@@ -112,7 +113,14 @@ class CallService(Capability):
         e_cb = partial(self._failure, cid, service)
 
         # Run service caller in the same thread.
-        ServiceCaller(trim_servicename(service), args, s_cb, e_cb, self.protocol.node_handle).run()
+        ServiceCaller(
+            trim_servicename(service),
+            args,
+            timeout,
+            s_cb,
+            e_cb,
+            self.protocol.node_handle,
+        ).run()
 
     def _success(self, cid, service, fragment_size, compression, message):
         outgoing_message = {
--- a/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/call_service.py
@@ -52,6 +52,12 @@ class CallService(Capability):
         # Call superclass constructor
         Capability.__init__(self, protocol)
 
+        self.default_timeout = (
+            protocol.node_handle.get_parameter("default_call_service_timeout")
+            .get_parameter_value()
+            .double_value
+        )
+
         # Register the operations that this capability provides
         call_services_in_new_thread = (
             protocol.node_handle.get_parameter("call_services_in_new_thread")
@@ -81,7 +87,7 @@ class CallService(Capability):
         fragment_size = message.get("fragment_size", None)
         compression = message.get("compression", "none")
         args = message.get("args", [])
-        timeout = message.get("timeout", 5.0)
+        timeout = message.get("timeout", self.default_timeout)
 
         if CallService.services_glob is not None and CallService.services_glob:
             self.protocol.log(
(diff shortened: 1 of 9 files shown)


rosbridge_library/src/rosbridge_library/capabilities/call_service.py
```diff
@@ -81,6 +81,7 @@ def call_service(self, message):
         fragment_size = message.get("fragment_size", None)
         compression = message.get("compression", "none")
         args = message.get("args", [])
+        timeout = message.get("timeout", 5.0)
```
sea-bass: So the previous default was to never time out. Should this continue to be the case?

rosbridge_library/src/rosbridge_library/internal/services.py
```diff
@@ -141,20 +152,29 @@ def call_service(
...
-    if not client.wait_for_service(server_timeout_time):
+    if not client.wait_for_service(server_ready_timeout):
         node_handle.destroy_client(client)
         raise InvalidServiceException(service)
 
     future = client.call_async(inst)
-    while rclpy.ok() and not future.done():
+    start_time = time.monotonic()
+    while (
+        rclpy.ok() and not future.done() and time.monotonic() - start_time < server_response_timeout
```
sea-bass: Is there no option to never time out? Like a -1 value or something?

bjsowa: We could treat any non-positive value as "no timeout"

rosbridge_library/src/rosbridge_library/capabilities/call_service.py
bjsowa: If we want to retain current behavior, it should by default never timeout.

bjsowa: Though I think we could change the defaults for rolling and future distros. @sea-bass what do you think about branching out for Jazzy ?

sea-bass: I'm OK either way.

It would be the "right" thing to do to branch out for Jazzy, but I think we still have a few more months where it's okay to land breaking changes to Jazzy.

As an unofficial rule, I would say we can keep putting changes into Jazzy until May 2025 where you may see more people upgrading from ROS 1.

rosbridge_library/src/rosbridge_library/internal/services.py
bjsowa: done

rosbridge_library/src/rosbridge_library/capabilities/call_service.py
bjsowa: I added `default_call_service_timeout` parameter. I think 5 seconds is a sane default value. I'll set it to 0 (no timeout) when backporting to Humble

rosbridge_library/test/capabilities/test_call_service.py
```diff
@@ -50,6 +58,12 @@ def setUp(self):
             self.trigger_cb,
             callback_group=self.cb_group,
         )
+        self.trigger_long_srv = self.node.create_service(
+            Trigger,
+            self.node.get_name() + "/trigger_long",
```
sea-bass, line 63: Can you just do `"trigger_long"` without the leading slash to get the same effect here?

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -178,6 +178,10 @@ def protocol_parameter_handling(self):
             "call_services_in_new_thread", False
         ).value
 
+        RosbridgeWebSocket.default_call_service_timeout = self.declare_parameter(
+            "default_call_service_timeout", 5.0
```
sea-bass, line 182: Should the default be `5.0` or indefinite wait?

sea-bass (commented):
I just have one question about what the default timeout should be -- for compatibility with the previous implementation, should it be a non-positive number for indefinite wait?

sea-bass (approved):
Wait, we already had this discussion about default timeout -- sorry, I'll approve.

rosbridge_server/scripts/rosbridge_websocket.py
bjsowa: The plan is to make it 5 seconds from Jazzy onward and indefinite when backporting it to Humble

rosbridge_library/test/capabilities/test_call_service.py
bjsowa: You mean without the node name as well? Like that
```suggestion
            "trigger_long",
```

PR #988, opened 2025-01-02
Add new service to retrieve the different interfaces in the ROS Network

FlorianLebecque:
**Public API Changes**

- Add a service to retrieve the different interfaces in the ROS network. Returned as an array of string

**Description**

Service : `/rosapi/interfaces`
Service type: `rosapi_msgs/srv/Interfaces`

```
---
string[] interfaces
```

This addition provides a new service that complements existing rosapi services. While rosapi allows retrieving topics, services, and their types, it lacked the ability to obtain a list of all available types. This new service addresses that gap.

### DIFF
--- a/rosapi/scripts/rosapi_node
+++ b/rosapi/scripts/rosapi_node
@@ -48,6 +48,7 @@ from rosapi_msgs.srv import (
     GetROSVersion,
     GetTime,
     HasParam,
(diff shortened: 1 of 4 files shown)


sea-bass (approved):
Looks good -- thanks!

Issue #769, opened 2022-08-01
Rosbridge blocks subscriptions to a topic if it subscribes before the subscription is available.


johncarl81:
**Description**

When I start rosbridge and subscribe to a topic before the topic is available I receive the following error:
```
[rosbridge_websocket]: New publisher discovered on topic '/mavros/local_position/pose', offering incompatible QoS. No messages will be received from it. Last incompatible policy: RELIABILITY
```
and subsequent attempts to connect via rosbridge to the given topic don't return anything.

- Library Version: 1.2.0-1focal.20220520.222649
- ROS Version: galactic
- Platform / OS: Docker - Ubuntu 20.04

**Steps To Reproduce**
1. Startup ros
```
ros2 daemon start
```
2. Startup rosbridge:
```
ros2 launch rosbridge_server rosbridge_websocket_launch.xml &
```
3. Subscribe to a stream that hasn't been started yet via ros (in this case a mavros stream):
(not sure of the exact syntax here):
subscribe to /mavros/global_position/global

4. Start said stream

5. Received error:
```
[rosbridge_websocket]: New publisher discovered on topic '/mavros/local_position/pose', offering incompatible QoS. No messages will be received from it. Last incompatible policy: RELIABILITY
```

**Expected Behavior**

I'd expect the subscription to be either configurable or more graceful in subscribing to this sort of stream.

**Actual Behavior**

In essence, rosbridge blocks itself from ever subscribing to a stream if rosbridge attempts to subscribe to it before it's available.  This block is in place for as long as rosbridge is continuously running.

**More Info**

Here's topic info for when rosbridge subscribes to the stream first:
Note the `Reliability: BEST_EFFORT` under `rosbridge_websocket`

```
ros2 topic info /mavros/global_position/global --verbose
Type: sensor_msgs/msg/NavSatFix

Publisher count: 1

Node name: global_position
Node namespace: /mavros
Topic type: sensor_msgs/msg/NavSatFix
Endpoint type: PUBLISHER
GID: b1.c4.10.01.32.d2.84.30.93.77.00.75.00.01.67.03.00.00.00.00.00.00.00.00
QoS profile:
Reliability: BEST_EFFORT
Durability: VOLATILE
Lifespan: 9223372036854775807 nanoseconds
Deadline: 9223372036854775807 nanoseconds
Liveliness: AUTOMATIC
Liveliness lease duration: 9223372036854775807 nanoseconds

Subscription count: 3

[...]

PR #991, opened 2025-01-09
Default subscriber QOS to BestEffort, account for TRANSIENT_LOCAL

ewak:
**Public API Changes**
Change default subscriber QosProfile from reliability=ReliabilityPolicy.RELIABLE to reliability=ReliabilityPolicy.BEST_EFFORT.

BUT if a publisher is found to have  durability = DurabilityPolicy.TRANSIENT_LOCAL, then use reliability=ReliabilityPolicy.RELIABLE

**Description**
IMO the current "sane" defaults for a subscriber are not great for unreliable links and/or large messages and are not compatible with publisher policy is not yet known.

Relates to https://github.com/RobotWebTools/rosbridge_suite/issues/769


bjsowa (changes requested):
That seems reasonable. Could you please add a parameter for the default subscription reliability policy? That way, it will be easier for me to backport it to Humble.

bjsowa (approved):

PR #992, opened 2025-01-14
Fix hardcoded services names in ROSAPI

FlorianLebecque:
**Public API Changes**
services name now have the namespace and node name as a prefix. If no namespace is defined, no change happens to the services
```
/rosapi/subscribers -> /namespace/nodename/subscribers
/rosapi/topic_type  -> /namespace/nodename/topic_type
/rosapi/topics      -> /namespace/nodename/topics
```

**Description**
Add missing namespace support for services in rosapi. Allowing more than one node running on the same network. This makes it easier to work with multi robot environment.

### DIFF
--- a/rosapi/scripts/rosapi_node
+++ b/rosapi/scripts/rosapi_node
@@ -89,47 +89,52 @@ class Rosapi(Node):
         else:
             full_name = self.get_namespace() + "/" + self.get_name()
         params.init(full_name)
-        self.create_service(Topics, "/rosapi/topics", self.get_topics)
-        self.create_service(Interfaces, "/rosapi/interfaces", self.get_interfaces)
-        self.create_service(TopicsForType, "/rosapi/topics_for_type", self.get_topics_for_type)
+
+        self.create_service(Topics, "~/topics", self.get_topics)
+        self.create_service(Interfaces, "~/interfaces", self.get_interfaces)
+        self.create_service(TopicsForType, "~/topics_for_type", self.get_topics_for_type)
         self.create_service(
             TopicsAndRawTypes,
-            "/rosapi/topics_and_raw_types",
+            "~/topics_and_raw_types",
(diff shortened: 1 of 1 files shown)


bjsowa:
Wouldn't it basically work the same if we just used topics in a private namespace? (e.g. `~/topics` or `~/interfaces`)

FlorianLebecque:
The services also need to be in a namespaces as we would have services collision if we run multiple node of rosapi

bjsowa:
> The services also need to be in a namespaces as we would have services collision if we run multiple node of rosapi

IIRC putting the name of the service in private namespace should prepend the node's name as well as it namespace. For example, `~/subscribers` should resolve to `/namespace/rosapi/subscribers`.

FlorianLebecque:
I've changed the modification, the rosapi services names were hardcoded. Now it should take into account the node name and namespace

bjsowa:
@FlorianLebecque This is not what I meant, just add `~` (tilde) character before topic name. For example, replace `full_name + "/topics"` with `~/topics`. Tilde is a special character that will be resolved by client libraries to node's private namespace.

FlorianLebecque:
Ok, yeah, I misunderstood. It should be ok now :)

bjsowa (approved):
I tested it and seems to work as expected. This breaks the current API though (if using a namespace) so if we want to backport it to Humble or Noetic this should be implemented as an opt-in feature. Jazzy is relatively new so we should be good with some breaking changes for a little while. @sea-bass What do you think?

sea-bass:
> I tested it and seems to work as expected. This breaks the current API though (if using a namespace) so if we want to backport it to Humble or Noetic this should be implemented as an opt-in feature. Jazzy is relatively new so we should be good with some breaking changes for a little while. @sea-bass What do you think?

I agree -- I think it's OK to have this go into Jazzy as is.

PR #994, opened 2025-01-20
Add timeout option to call_service messages (backport #984)

bjsowa:


rosbridge_library/test/capabilities/test_service_capabilities.py
```diff
@@ -141,18 +140,73 @@ def test_call_advertised_service(self):
...
-            loop_iterations += 1
-            if loop_iterations > 3:
+            if time.monotonic() - start_time > 0.3:
                 self.fail("Timed out waiting for service response message.")
 
         self.assertFalse(self.received_message is None)
         self.assertEqual(self.received_message["op"], "service_response")
         self.assertTrue(self.received_message["result"])
 
+    @unittest.skip("Gets stuck in rclpy.shutdown on teardown")
```
bjsowa, line 153: I have no idea why this test gets stuck, tried different things but nothing worked. Probably a bug in rclpy

sea-bass: If you use a `MultiThreadedExecutor` instead of the (single-threaded) global ROS executor, does it work?

Or even better, if you set:

```python
self.node.declare_parameter("call_services_in_new_thread", True)
```

sea-bass (approved):

PR #996, opened 2025-01-23
Call services and send action goals in new threads by default

bjsowa:
**Public API Changes**
None

**Description**
Since we've already made some breaking changes for Jazzy, I would also like to change the default value of `call_services_in_new_thread` parameter to true. IMO it gives better out-of-box experience.


sea-bass (commented):
Good idea. Can you do the same for the actions parameter?

sea-bass (approved):
Thanks! Just rename the PR title to include them both and LGTM

PR #997, opened 2025-01-23
Add warnings about default parameter values changes

bjsowa:
**Public API Changes**
None

**Description**
Related to #996


sea-bass (commented):
Same comment on doing the same for actions / renaming the PR

bjsowa:
@sea-bass Not sure if this is excessive spam of annoying warnings or not.

sea-bass (approved):
I think it's fine! Deprecations are best handled when the warnings are annoying

PR #1001, opened 2025-01-27
New async rosapi params module implementation

bjsowa:
**Public API Changes**
The `set_param`, `get_param`, and `delete_param` service types now include:
- A boolean indicating whether the operation was successful.
- An optional string explaining the reason for failure.

These changes do not break the compatibility with the current version of `roslibjs` client, as it will ignore these new fields. Additionally, to make it backward compatible, if `get_param` fails, the value field in the response will still be set to an empty JSON string.

**Description**
Changed the implementation of rosapi params module to take advantage of async/await feature of rclpy executor instead of creating a separate node.

Aside from the aforementioned API change, there are few implementation changes:
- `get_param_names` calls `list_parameters` service for each node in parallel, not sequentially. This should make it much more robust, especially if the node graph is quite big.
- all rosapi param services use `ReentrantCallbackGroup` which means multiple service callbacks can be processed asynchronously.
- The helper functions from `ros2param` cli package are no longer used as they are not awaitable.


rosapi/src/rosapi/params.py
```diff
@@ -31,27 +32,26 @@
...
+from rcl_interfaces.srv import GetParameters, ListParameters, SetParameters
+from rclpy.callback_groups import MutuallyExclusiveCallbackGroup
+from rclpy.node import Node
 from rclpy.parameter import get_parameter_value
+from rclpy.task import Future
 from ros2node.api import get_absolute_node_name
-from ros2param.api import call_get_parameters, call_set_parameters
 from rosapi.proxy import get_nodes
 
+from .async_helper import futures_wait_for
```
sea-bass: Can you use the fully qualified instead of relative path?

rosapi/src/rosapi/params.py
```diff
@@ -128,47 +118,93 @@ def _set_param(node_name, name, value, parameter_type=None):
...
-    try:
-        # call_get_parameters will fail if node does not exist.
-        call_set_parameters(node=_node, node_name=node_name, parameters=[parameter])
-    except Exception:
-        pass
+    assert _node is not None
+    client = _node.create_client(
+        SetParameters,
+        f"{node_name}/set_parameters",
+        callback_group=MutuallyExclusiveCallbackGroup(),
```
sea-bass, line 126: Did you also want this to be reentrant callback group?

rosapi/src/rosapi/params.py
```diff
@@ -128,47 +118,93 @@ def _set_param(node_name, name, value, parameter_type=None):
...
 
-def has_param(node_name, name, params_glob):
+async def _get_param(node_name: str, name: str) -> ParameterValue:
+    """Internal helper function for get_param"""
+
+    assert _node is not None
+    client = _node.create_client(
+        GetParameters,
+        f"{node_name}/get_parameters",
+        callback_group=MutuallyExclusiveCallbackGroup(),
```
sea-bass, line 177: Same here and onwards

rosapi_msgs/srv/GetParam.srv
```diff
@@ -1,4 +1,5 @@
 string name
-string default_value
```
sea-bass, line 2: The PR says that the extra response fields are non-intrusive as you can just ignore them, but removing this default value probably will affect a few users.

I think it makes sense to remove it since you're just getting a parameter, and can handle falling back to a default on the client side, but just pointing out this is another breakage.

sea-bass (approved):
Overall looks good -- no major comments

rosapi/src/rosapi/params.py
bjsowa: No need for service clients as they are created only for a single service call. In fact, if I understand it correctly, using the default callback group should be also ok here but, just in case I'm wrong, I create unique callback groups for each client.

rosapi/src/rosapi/params.py
bjsowa: done

rosapi_msgs/srv/GetParam.srv
bjsowa: AFAIK roslibjs does not use this feature but you're right that it still may affect some users, so I restored this field.

Issue #990, opened 2025-01-06
After running for a period of time, a certain topic may not be subscribed to


nvidia-tzkj:
Recently, when using rosbridge_server, I found that occasionally there may be situations where a certain topic is not subscribed. We use several different clients to establish connections with it. Other status subscriptions are normal, but robot_pose displays subscribed but does not return data. This topic currently has the highest frequency among all topics, at 20Hz

- Library Version: 
- ROS Version:  ROS1
- Platform / OS: ubuntu 20.04
this is our param setting:

- arg name="port" default="9090"
- arg name="address" default="0.0.0.0" 
- arg name="ssl" default="false" 
- arg name="certfile" default=""
- arg name="keyfile" default="" 
- arg name="retry_startup_delay" default="5" 
- arg name="fragment_timeout" default="600" 
- arg name="delay_between_messages" default="0" 
- arg name="max_message_size" default="None"  
- arg name="unregister_timeout" default="10000000" 
- arg name="websocket_external_port" default="None" 
- arg name="use_compression" default="false" 
- arg name="authenticate" default="false" 
- arg name="websocket_ping_interval" default="0" 
- arg name="websocket_ping_timeout" default="30" 
- arg name="websocket_null_origin" default="true" 
- arg name="topics_glob" default="[*]" 
- arg name="services_glob" default="[*]" 
- arg name="params_glob" default="[*]" 
- arg name="bson_only_mode" default="false" 
- arg name="output" default="screen" 
<node name="rosbridge_websocket" pkg="rosbridge_server" type="rosbridge_websocket" output="$(arg output)">
      <param name="authenticate" value="$(arg authenticate)" />
      <param name="port" value="$(arg port)"/>
      <param name="address" value="$(arg address)"/>
      <param name="retry_startup_delay" value="$(arg retry_startup_delay)"/>
      <param name="fragment_timeout" value="$(arg fragment_timeout)"/>
      <param name="delay_between_messages" value="$(arg delay_between_messages)"/>
      <param name="max_message_size" value="$(arg max_message_size)"/>
      <param name="unregister_timeout" value="$(arg unregister_timeout)"/>
      <param name="use_compression" value="$(arg use_compression)"/>

      <param name="websocket_ping_interval" value="$(arg websocket_ping_interval)" />
      <param name="websocket_ping_timeout" value="$(arg websocket_ping_timeout)" />
      <param name="websocket_external_port" value="$(arg websocket_external_port)" />

[...]

PR #1004, opened 2025-02-06
catch race condition if websocket is closed while messages are in queue

jobafr:
**Public API Changes**
None

**Description**
Running rosbridge against newer versions of autobahn introduces a race condition. If a websocket client disconnects while the server still tries to send something to it, an `autobahn.exceptions.Disconnect` will be raised. This breaks all subsequent communication on that topic (or service, i believe) for any client.

This PR changes the behavior so that the Disconnected exception is ignored. The client will still be cleaned up properly shortly after the failed send.

The behavior was apparently changed in autobahn 20.1.3:
https://github.com/crossbario/autobahn-python/pull/1299
https://autobahn.readthedocs.io/en/latest/changelog.html

- It looks like other people have run into this https://github.com/RobotWebTools/rosbridge_suite/issues/990 (opened by @nvidia-tzkj last month).

As far as I can tell, the ros2 branch has proper error handling in place already, but the code structure is somewhat different:
https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/rosbridge_server/src/rosbridge_server/websocket_handler.py#L190

### DIFF
--- a/rosbridge_server/src/rosbridge_server/autobahn_websocket.py
+++ b/rosbridge_server/src/rosbridge_server/autobahn_websocket.py
@@ -42,6 +42,21 @@ from functools import wraps
 from collections import deque
 
 from autobahn.twisted.websocket import WebSocketServerProtocol
+
+SendException = Exception
+try:
+    # Newer autobahn versions (>= 20.1.3) raise an exception when sending to a closed socket
+    # PR: https://github.com/crossbario/autobahn-python/pull/1299
+    # Changelog: https://autobahn.readthedocs.io/en/latest/changelog.html
+    from autobahn.exception import Disconnected
+    SendException = Disconnected
+except ImportError:
+    # Older versions fail silently, so create a fake exception we can pretend to catch
+    # so we can ignore the error.
+    class NeverException(Exception):
+        pass
+    SendException = NeverException
+
 from twisted.internet import interfaces, reactor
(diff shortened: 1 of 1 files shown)


bjsowa:
I fixed the noetic workflow. Could you rebase your PR or merge the `ros1` branch to run the workflow again?

jobafr:
I rebased it against your ros1 branch.

bjsowa:
Can this be ported to `ros2` branch or does it only apply to code structure in `ros1` branch?

bjsowa:
> Can this be ported to `ros2` branch or does it only apply to code structure in `ros1` branch?

Nevermind, I know see that the ros2 implementation has diverged just before ros1 branch changed websocket implementation to autobahn and the change was never ported. I'm wondering if it would be a good idea to port this change at this point. @sea-bass What do you think?

sea-bass:
> Nevermind, I know see that the ros2 implementation has diverged just before ros1 branch changed websocket implementation to autobahn and the change was never ported. I'm wondering if it would be a good idea to port this change at this point. @sea-bass What do you think?

I think yes, but only for rolling?

bjsowa:
> I think yes, but only for rolling?

Sure, I might do this if I'll find the time.

bjsowa (approved):
LGTM

Issue #1002, opened 2025-02-04
'ascii' codec can't decode byte


alien19:
**Description**
I'm using rosbridge_server to publish some rostopics data in bson_only mode. The messages are of type sensor_msgs/Imu and sensor_msgs/Image.
I can send the subscribe message to the server in bson format, in case of the imu topic the client gets the data correctly but in case of Image topic the server gives me \
Exception calling subscribe callback: 'ascii' codec can't decode byte 0xff in position 0: ordinal not in range(128)

- Library Version: newest version (0.11.17-1focal.20240913.194148)
- ROS Version: Noetic
- Platform / OS: Ubuntu 20.04 (WSL)

**Steps To Reproduce**
In Python: \
on the client side: 
```
import websocket
import bson
import cbor2


def on_message(ws, message):
    print("Received message:", bson.BSON.decode(message))

def on_error(ws, error):
    print("Error:", error)

def on_close(ws, close_status_code , close_msg):
    ws.close
    print("Connection closed", close_status_code, close_msg)

def on_open(ws):
    # Create a BSON message
    message = {
        "op": "subscribe",
        "topic": "/mcsrov/imu"
        }
    ws.send_bytes(bson.BSON.encode(message))

if __name__ == "__main__":
    websocket.enableTrace(True)
    ws = websocket.WebSocketApp("ws://localhost:9080",
                                on_message=on_message,
                                on_error=on_error,
                                on_close=on_close)
    ws.on_open = on_open
    ws.run_forever()
```
\
on the server side, in a launch file:
```
<include file="$(find rosbridge_server)/launch/rosbridge_websocket.launch" > 
    <arg name="port" value="9080"/>
    <arg name="bson_only_mode" value="true"/>
  </include>
```

**Expected Behavior**
the client gets the image message binary-encoded

**Actual Behavior**
no Image msg received

alien19:
**Update**
I have reached that the line [208](https://github.com/RobotWebTools/rosbridge_suite/blob/0d96ae5816bf6c4aebe86d9ff6d4950d5d2768e8/rosbridge_library/src/rosbridge_library/internal/message_conversion.py#L208) is causing the error where the uint8[] in ros is dealt as string which is not the case so, I see no use of encoding then decoding just returning the "inst" is enough, I'm working on it and will check its efficiency

PR #1006, opened 2025-02-06
Deal with uint8[] as binary data not string

alien19:
**Public API Changes**

None

**Description**

the change made, allows uint8[] raw data to be sent as it is (binary) in the bson_only mode

solves #1002


bjsowa:
Please fix the failing test.

PR #1007, opened 2025-02-06
Update setup-ros and action-ros-ci actions in CI workflow

bjsowa:
Seems to be enough to fix CI workflow


bjsowa:
@sea-bass I would have probably merged it without review but I don't have enough permissions to omit branch protection rules.

sea-bass (approved):

PR #1008, opened 2025-02-12
Fix randomly failing subscribe unsubscribe test

bjsowa:
This test keeps randomly failing in Github actions and it's starting to drive me nuts


bjsowa:
I've run the workflow 10 times and it didn't fail a single time now.

sea-bass (approved):

PR #1013, opened 2025-02-26
Fix action cancelling/aborting

bjsowa:
**Public API Changes**
None

**Description**
I kept getting frustrated by a failing action test so I spent some time to investigate and found 2 problems with the implementation:
1. When cancelling the goal before the action client has received the goal handle, it does not get cancelled.
2. When unadvertising the action when goal is executing, the goal does not get aborted.

This PR attemps to fix these 2 issues. I also made the test run 3 times faster.


sea-bass (approved):
Wonderful -- thanks!

PR #1018, opened 2025-03-19
Fix array-like parameter serialization in rosbridge get_param

noah-wardlow:
This PR enhances the parameter value handling in the get_param function to ensure proper JSON serialization of array-like data types. Previously, when retrieving parameters that contained array-like values (such as numpy arrays, Python arrays, or simple numeric values that should be treated as single-element arrays), they might not be properly serialized to JSON format, resulting in empty strings or invalid representations.

The changes include:
- Adding proper array type detection and conversion prior to JSON serialization
- Converting objects with a tolist() method (e.g., numpy arrays) to Python lists
- Converting Python array objects to regular lists

This fix improves interoperability between ROS parameters and JSON-based web interfaces, ensuring that parameter values are consistently and correctly serialized, especially for numeric array parameters commonly used in robotics applications.

The issue can be reproduced by calling a get_param for a a vector value. Previously the value would come back as an empty string:

``` 
# Before fix - parameter retrieval via rosbridge
   $ ros2 param get /velocity_force_controller max_cartesian_velocity
   Double values are: array('d', [0.25, 0.25, 0.25, 1.5707, 1.5707, 1.5707])
   
   # But when accessed through rosbridge API:
   Value: ""
```   

After this PR, the parameter is properly converted to a JSON serializable format:

```
# After fix - parameter retrieval via rosbridge
$ ros2 param get /velocity_force_controller max_cartesian_velocity
Double values are: array('d', [0.25, 0.25, 0.25, 1.5707, 1.5707, 1.5707])

# When accessed through rosbridge API:
Value: [0.25, 0.25, 0.25, 1.5707, 1.5707, 1.5707]
```

### DIFF
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -31,6 +31,7 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import array
 import fnmatch
 from json import dumps, loads
 
@@ -163,6 +164,12 @@ async def get_param(node_name: str, name: str, params_glob: str) -> str:
     node_name = get_absolute_node_name(node_name)
     pvalue = await _get_param(node_name, name)
     value = getattr(pvalue, _parameter_type_mapping[pvalue.type])
+
+    # Convert array types to lists for JSON serialization
+    if hasattr(value, "tolist"):  # This will catch numpy arrays and Python arrays
+        value = value.tolist()
+    elif isinstance(value, array.array):
+        value = list(value)
 
     return dumps(value)
 
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -31,7 +31,6 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import array
 import fnmatch
 from json import dumps, loads
 
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -168,8 +168,6 @@ async def get_param(node_name: str, name: str, params_glob: str) -> str:
     # Convert array types to lists for JSON serialization
     if hasattr(value, "tolist"):  # This will catch numpy arrays and Python arrays
         value = value.tolist()
-    elif isinstance(value, array.array):
-        value = list(value)
 
     return dumps(value)
 


rosapi/src/rosapi/params.py
```diff
@@ -164,6 +165,14 @@ async def get_param(node_name: str, name: str, params_glob: str) -> str:
     pvalue = await _get_param(node_name, name)
     value = getattr(pvalue, _parameter_type_mapping[pvalue.type])
 
+    # Convert array types to lists for JSON serialization
+    if hasattr(value, "tolist"):  # This will catch numpy arrays and Python arrays
+        value = value.tolist()
+    elif isinstance(value, (array.array, list)):
+        value = list(value)
+    elif isinstance(value, (float, int)):
```
sea-bass: you could consider `numbers.Number` here: https://stackoverflow.com/a/4187266

sea-bass (commented):
o hey @noah-wardlow!

rosapi/src/rosapi/params.py
```diff
@@ -164,6 +165,14 @@ async def get_param(node_name: str, name: str, params_glob: str) -> str:
     pvalue = await _get_param(node_name, name)
     value = getattr(pvalue, _parameter_type_mapping[pvalue.type])
 
+    # Convert array types to lists for JSON serialization
+    if hasattr(value, "tolist"):  # This will catch numpy arrays and Python arrays
+        value = value.tolist()
+    elif isinstance(value, (array.array, list)):
```
sea-bass: if the value is already a `list`, you can just return is as is, right? No need to cast it

noah-wardlow: Hey @sea-bass! good point here, just updated

rosapi/src/rosapi/params.py
noah-wardlow: I think this line was unnecessary actually, I just took it out. Scalar numbers appear to parse fine and don't need to be converted

sea-bass (approved):

bjsowa:
@noah-wardlow I was not aware ROS parameters can be deserialized into anything other than a list in Python. Do you any example code that declares a parameter in such way?

noah-wardlow:
> @noah-wardlow I was not aware ROS parameters can be deserialized into anything other than a list in Python. Do you any example code that declares a parameter in such way?

As far as I understand, support was added here: https://github.com/ros2/rclpy/pull/444

and these specialized array types don't natively serialize to JSON. 


Specifically in our code base we have

```
max_cartesian_velocity: {
  type: double_array,
  description: "Cartesian velocity limits (absolute value), one element per Cartesian axis (x, y, z, rx, ry, rz).",
  validation: {
    fixed_size<>: 6,
    lower_element_bounds<>: 0.001
  }
}
```

In the C++ controller, this parameter is mapped to an Eigen vector

```
.cartesian_velocity_limits =
    Eigen::Map<const Eigen::Vector6d>(params_.max_cartesian_velocity.data(), params_.max_cartesian_velocity.size()),
   
 ```

@bjsowa

File: rosapi/src/rosapi/params.py [hunk 3]
```diff
@@ -164,6 +165,12 @@ async def get_param(node_name: str, name: str, params_glob: str) -> str:
     pvalue = await _get_param(node_name, name)
     value = getattr(pvalue, _parameter_type_mapping[pvalue.type])
 
+    # Convert array types to lists for JSON serialization
+    if hasattr(value, "tolist"):  # This will catch numpy arrays and Python arrays
+        value = value.tolist()
+    elif isinstance(value, array.array):
+        value = list(value)
```
bjsowa: If the first conditional will already work for Python arrays, why do we need the second one?
```suggestion
```

noah-wardlow: Yeah good point. I think I was trying to be defensive against non standard list types but I don't think it's actually applicable

rosapi/src/rosapi/params.py
```diff
@@ -31,6 +31,7 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import array
```
bjsowa: This is redundant now
```suggestion
```

bjsowa (changes requested):

bjsowa (approved):
I looked at the code generated by `rosidl_generator_py` for `ParameterValue` messages and it looks like `integer_array_value` and `double_array_value` are deserialized into `array.array` (contrary to [ROS docs](https://docs.ros.org/en/rolling/Concepts/Basic/About-Interfaces.html)), so this change makes sense.

PR #1032, opened 2025-05-16
Add subscription ID tracking in message publishing

Josme:
Description :
This PR implements subscription ID propagation in message publishing to enable better client-side message tracking. The main changes include:

1. Added subscription ID (sid) injection for all message formats:
```
for sid in subscription.clients:
    outgoing_msg = {"op": "publish", 
    "topic": topic, "id": sid}
```
2. Enhanced compression format handling with ID preservation:
```
# For PNG compression
outgoing_msg = {"op": "png", "data": 
encode_png(...), "id": sid}

# For CBOR formats
outgoing_msg["id"] = sid
```
3. Implemented per-client message generation logic:
```
subscription = self._subscriptions.get
(topic)
if subscription:
    for sid in subscription.clients:
        # Message construction per client
```


bjsowa:
I like the premise but I think I have to drop this feature as it could break existing applications. When the client subscribes many times to the same topic (which is a common case), the protocol will now send each message multiple times which could overload the network.

PR #1046, opened 2025-06-15
 Added ActionType service and related functionality

chaitanya-mandala:
- Introduced ActionType service to retrieve the type of a specified ROS action.
- Updated proxy to include logic for determining action types based on existing action servers.
- Modified rosapi_node to create the new ActionType service endpoint.
- Added ActionType message definition in rosapi_msgs.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,7 +7,7 @@
     "[python]": {
         "editor.formatOnSave": true,
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "explicit"
         }
     },
     "python.linting.banditArgs": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,7 +7,7 @@
     "[python]": {
         "editor.formatOnSave": true,
         "editor.codeActionsOnSave": {
-            "source.organizeImports": "explicit"
+            "source.organizeImports": true
         }
     },
     "python.linting.banditArgs": [
--- a/rosapi/package.xml
+++ b/rosapi/package.xml
@@ -25,6 +25,7 @@
   <exec_depend>rosbridge_library</exec_depend>
   <exec_depend>ros2node</exec_depend>
   <exec_depend>ros2service</exec_depend>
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
```diff
@@ -7,7 +7,7 @@
     "[python]": {
         "editor.formatOnSave": true,
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "explicit"
```
bjsowa: Please revert this change

bjsowa (changes requested):

bjsowa:
@chaitanya-mandala Why not use `_node.get_action_names_and_types()` instead of identifying actions by hidden topics?

chaitanya-mandala:
> @chaitanya-mandala Why not use `_node.get_action_names_and_types()` instead of identifying actions by hidden topics?


@bjsowa Thank you for the insightful suggestion! I've updated the get_action_type() function to utilize the official graph call from the ros2action library and the functionality remains consistent.

Here is test Result
```bash
chaitu@chaitu:~$ ros2 service call /rosapi/action_type rosapi_msgs/srv/ActionType '{action: "/fibonacci"}'
requester: making request: rosapi_msgs.srv.ActionType_Request(action='/fibonacci')

response:
rosapi_msgs.srv.ActionType_Response(type='action_tutorials_interfaces/action/Fibonacci')
```

bjsowa (changes requested):
Looks much simpler this way. Personally, I would rather use the method from `rclpy` directly instead of wrappers from `ros2cli` packages but since other proxy methods are also implemented this way, we can leave this and change in future code refactor.

The only change request I have is to add `ros2action` to package dependencies

chaitanya-mandala:
> Looks much simpler this way. Personally, I would rather use the method from `rclpy` directly instead of wrappers from `ros2cli` packages but since other proxy methods are also implemented this way, we can leave this and change in future code refactor.
> 
> The only change request I have is to add `ros2action` to package dependencies

Thank you for your feedback! I've added ros2action dependency in the package.xml.

bjsowa (approved):

PR #1040, opened 2025-06-04
fix: Reduce idle CPU consumption of websocket server

bjsowa:
**Public API Changes**
None

**Description**
This PR removes the use of [tornado.ioloop](https://www.tornadoweb.org/en/stable/ioloop.html) and instead uses the `asyncio` event loop interface, as recommended in tornado documentation since version 6.0.

The ROS spinning is now done in a separate thread instead of a periodic async callback.

SIGINT and SIGTERM signal handling is now done manually to ensure graceful shutdown.

In my case, the idle CPU consumption went from 9% to 0%.
With a client connected, the consumption went from 38% to 34%.

This is still quite high but a good step forward. I tried improving it further by using [EventsExecutor](https://github.com/ros2/rclpy/issues/1389) and the consumption with the client connected went from 34% to 18% which is huge but I'm reluctant to use it as it is experimental.

PR #981, opened 2024-12-18
Drop support for ROS 2 Iron

sea-bass:
**Public API Changes**
None

**Description**
This PR removes support for ROS 2 Iron now that it's EOL.

Additionally, this lets us re-enable some unit tests that were previously failing on Iron.

PR #1048, opened 2025-07-01
fix: rosbridge_websocket cooperative shutdown

cyan-at:
**Public API Changes**
None

**Description**
We use this package in a systemctl
When we do `systemctl ... stop <our.service>`, a common thing blocking shutdown is this process
We can work around it, and locally do `systemctl ... stop --force` but this patch isn't global and not ideal

I tested some local changes to this code to make it cooperate, without any obvious regressions

I made no issues, but maybe one exists or I can make one for tracking

### DIFF
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -339,10 +339,13 @@ def main(args=None):
     rclpy.init(args=args)
     node = RosbridgeWebsocketNode()
 
-    executor = rclpy.executors.MultiThreadedExecutor()
+    executor = rclpy.executors.SingleThreadedExecutor()
     executor.add_node(node)
 
     def spin_ros():
+        if not rclpy.ok():
+            shutdown_hook()
+            return
         executor.spin_once(timeout_sec=0.01)
         if not rclpy.ok():
             shutdown_hook()
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -347,8 +347,6 @@ def main(args=None):
             shutdown_hook()
             return
         executor.spin_once(timeout_sec=0.01)
-        if not rclpy.ok():
-            shutdown_hook()
 
     spin_callback = PeriodicCallback(spin_ros, 1)
     spin_callback.start()
@@ -358,8 +356,8 @@ def main(args=None):
         rclpy.shutdown()
     except KeyboardInterrupt:
         print("Exiting due to SIGINT")
-        spin_callback.stop()
     finally:
+        spin_callback.stop()
         shutdown_hook()  # shutdown hook to stop the server
 
 
--- a/rosbridge_server/scripts/rosbridge_websocket.py
+++ b/rosbridge_server/scripts/rosbridge_websocket.py
@@ -339,7 +339,7 @@ def main(args=None):
     rclpy.init(args=args)
     node = RosbridgeWebsocketNode()
 
-    executor = rclpy.executors.SingleThreadedExecutor()
+    executor = rclpy.executors.MultiThreadedExecutor()
     executor.add_node(node)
 
     def spin_ros():
@@ -355,6 +355,7 @@ def main(args=None):
         rclpy.shutdown()
     except KeyboardInterrupt:
         print("Exiting due to SIGINT")
+        spin_callback.stop()
     finally:
         shutdown_hook()  # shutdown hook to stop the server
 


cyan-at:
@bjsowa can you give this a look? thanks

bjsowa:
Is switching to MultiThreadedExecutor necessary? I believe we are trying to avoid it due to some existing bugs that may lead to a deadlock. (@sea-bass or @MatthijsBurgh might know more).

I think I got cooperative shutdown to work well in #1040 but it is not passing tests due to bugs in rclpy which I am trying to fix but couldn't find time to finish.

sea-bass:
`MultiThreadedExecutor` has a lot of issues and I would avoid it unless absolutely necessary. But two suggestions:

1. This script in question already has a bunch of command-line args. So maybe one that allows switching executors isn't a bad idea. (Side question, this really shouldn't use raw `sys.argv` but rather something like `argparse`, though I understand it gets in the way of `ros2`)
2. Could always try the new `EventsExecutor` which is meant to significantly help with CPU usage? But I believe it's just now an experimental Python executor in Kilted.

cyan-at:
Thanks all for the feedback

MultiThreadExecutor, + [ReentrantCallbackGroup](https://docs.ros.org/en/foxy/How-To-Guides/Using-callback-groups.html#basics-of-callback-groups) might help with some of that deadlock (handwavy gestures at threading.RLock)

But that indicates changes *inside* the node wherever it has such a mutexed object

It does seem necessary, without it there is some circular dependency
Maybe I can try some different ordering or something else

But if this is going to be throwaway in light of #1040, then we can put this to bed in favor of that and it's other benefits

MatthijsBurgh:
See https://github.com/RobotWebTools/rosbridge_suite/pull/981 and https://github.com/ros2/rclpy/issues/1223. This is probably still an issue.
Using the `MultiThreadedExecutor` is the issue. It is not related to deadlocks. But performance related.

MatthijsBurgh:
@sea-bass There are no problems with using `argparse` as long as you combine it with the `remove_ros_args` function from `rclpy`, see https://docs.ros.org/en/iron/p/rclpy/rclpy.utilities.html#rclpy.utilities.remove_ros_args

MatthijsBurgh:
The `EventExecutor` looks interesting. But don't enforce it for now.

Though I think the entire executor change in this PR is not needed to fix the shutdown issues.

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -339,7 +339,7 @@ def main(args=None):
     rclpy.init(args=args)
     node = RosbridgeWebsocketNode()
 
-    executor = rclpy.executors.SingleThreadedExecutor()
+    executor = rclpy.executors.MultiThreadedExecutor()
```
MatthijsBurgh: Get rid of this change as it is not needed to fix the shutdown issues.

MatthijsBurgh (changes requested):

sea-bass:
> @sea-bass There are no problems with using `argparse` as long as you combine it with the `remove_ros_args` function from `rclpy`, see https://docs.ros.org/en/iron/p/rclpy/rclpy.utilities.html#rclpy.utilities.remove_ros_args

Whoa, I did not know about this. Thank you!

rosbridge_server/scripts/rosbridge_websocket.py
cyan-at: hmmm

in our system this is necessary

when the systemctl shutdown happens, not ok message below doesn't appear 

```
        executor.spin_once(timeout_sec=0.01)
        node.get_logger().warn("tick")
        if not rclpy.ok():
            node.get_logger().warn("not ok!")
            shutdown_hook()  
```

It suggests the timeout_sec isn't being working as intended, maybe the rclpy context is dead and spin_once blocks forever unless it is a MultiThreaded?

cyan-at: @MatthijsBurgh 

I added a ok() check before spinning, and that worked, it indicates my suspicion that external systemd shutdown kills the context and that makes singlethreadexecutor spinning block forever, but I can't explicitly confirm it

Resolving this

cyan-at:
Hi @bjsowa @MatthijsBurgh @sea-bass , I got it to work with SingleThreadExecutor

If you think the ok() check after spin_once is redundant, I can remove it

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -355,6 +358,7 @@ def spin_ros():
         rclpy.shutdown()
     except KeyboardInterrupt:
         print("Exiting due to SIGINT")
+        spin_callback.stop()
```
MatthijsBurgh: I would say, you also need to stop the `spin_callback` in case of any other exception. Correct?

rosbridge_server/scripts/rosbridge_websocket.py
```diff
@@ -343,6 +343,9 @@ def main(args=None):
     executor.add_node(node)
 
     def spin_ros():
+        if not rclpy.ok():
+            shutdown_hook()
+            return
         executor.spin_once(timeout_sec=0.01)
         if not rclpy.ok():
             shutdown_hook()
```
MatthijsBurgh: This one seems redundant indeed

MatthijsBurgh (changes requested):
Would be possible to create a test for your case? So we have a test for a robust shutdown.

rosbridge_server/scripts/rosbridge_websocket.py
cyan-at: removed the one after

it is common form to check for a hb / flag *before* doing work

rosbridge_server/scripts/rosbridge_websocket.py
cyan-at: i've moved this to the `finally` block

MatthijsBurgh:
@cyan-at Would be possible to create a test for your case? So we have a test for a robust shutdown.

cyan-at:
@MatthijsBurgh  I was looking at the existing test suites

I saw a lot of `make_client` but didn't see a good way to exercise this kind of shutdown in the python `unittest` context

Do you have any suggestions?

cyan-at:
@MatthijsBurgh I don't think this change warrants a test, and I worry adding a test might introduce other changes out of scope

I can provide a test report i.e. before vs after behavior in this PR and I hope that would be sufficient

MatthijsBurgh (approved):

MatthijsBurgh:
Waiting for a final review of @bjsowa. You can merge it if you approve

bjsowa (approved):

cyan-at:
Thanks all!

PR #1053, opened 2025-07-27
Clean up package dependencies

bjsowa:
**Public API Changes**
None

**Description**
This PR fixes some issues with the dependency declaration. Most notably:
- Adds missing dependencies and removes unused ones,
- Uses correct dependency types, for example, python code does not need build dependencies and for dependencies of rosidl interfaces we should be using the `<depend>` tags which implies `build_depend`, `exec_depend` and `build_export_depend`
- (Opinionated) Prefers `ament_cmake` over `ament_cmake_ros`. The only exception is `rosbridge_library` which is now using the `ament_add_ros_isolated_pytest_test` macro from `ament_cmake_ros`
- (Opinionated) Sorts the dependencies in lexicographic order (easier to avoid duplicates and merge multiple code changes in the future)

I split the changes into many small commits to make it easy to follow.

### DIFF
--- a/rosapi/CMakeLists.txt
+++ b/rosapi/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 project(rosapi)
 
-find_package(ament_cmake_ros REQUIRED)
+find_package(ament_cmake REQUIRED)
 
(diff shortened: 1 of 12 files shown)


sea-bass (approved):
Nice!

PR #1059, opened 2025-08-08
refactor: Enable various ruff checks and fix lint errors

bjsowa:
I enabled as many checks as I could without rewriting too much and added explanations on why some checks are disabled.

I decided on `max-line-length` of 120, as setting it to 100 would be too exhausting to fix.


EzraBrooks (approved):
LGTM since tests are passing. This is a great selection of rules to enable!

bjsowa:
I'll wait for one other review since this changes a lot of code

rosapi/scripts/rosapi_node
```diff
@@ -458,7 +458,7 @@ class Rosapi(Node):
                     default = loads(request.default_value)
                 except ValueError:
                     self.get_logger().error(
-                        "Failed to parse default value: {}".format(request.default_value)
+                        f"Failed to parse default value: {request.default_value}"
```
sea-bass, line 461: aw yea, f-strings

rosapi/src/rosapi/params.py
```diff
@@ -72,7 +73,7 @@
 ]
 
 
-def init(node: Node, timeout_sec: float | int = DEFAULT_PARAM_TIMEOUT_SEC):
+def init(node: Node, timeout_sec: float = DEFAULT_PARAM_TIMEOUT_SEC):
```
sea-bass, line 76: I believe this one should remain `float | int` based on the code, right?

rosbridge_library/src/rosbridge_library/internal/subscription_modifiers.py
```diff
@@ -75,10 +75,9 @@ def handle_message(self, msg):
     def transition(self):
         if self.throttle_rate == 0 and self.queue_length == 0:
             return self
-        elif self.queue_length == 0:
+        if self.queue_length == 0:
             return ThrottleMessageHandler(self)
-        else:
-            return QueueMessageHandler(self)
+        return QueueMessageHandler(self)
```
sea-bass: I feel like this code could be simplified to not have 2 of the same check for queue length. Something like

```python
if self.queue_length > 0:
    return QueueMessageHander(self)
if self.throttle_rate > 0:
    return ThrottleMessageHandler(self)
return self

sea-bass: similar stuff extends to the rest of this file, but I guess another PR would be better at this point... or just doing nothing, it's such a small thing.

sea-bass (approved):

rosapi/src/rosapi/params.py
bjsowa: https://docs.astral.sh/ruff/rules/redundant-numeric-union/

sea-bass: oh, interesting!

rosbridge_library/src/rosbridge_library/internal/subscription_modifiers.py
bjsowa: Seems more readable that way. I've rewritten all transitions. Could you recheck?

sea-bass: Looks correct -- thank you!

PR #1060, opened 2025-08-08
feat: New parameter and cli arguments handling 

bjsowa:
Let's be honest here. The way parameters are passed down from rosbridge_server to the Rosbridge protocol and internal classes is a complete mess:
- Rosbridge library retrieves some parameters either from:
  1. `Protocol.parameters` dictionary.
  2. Node handle parameters.
  3. Class variables overwritten by rosbridge_server.
- RosbridgeWebsocketNode parameter handling is very complex. It's very hard to tell which parameters relate to the server and which are passed to the protocol.
- The cli argument parsing contains a lot of repeated boilerplate code.
- Some internal modules like `publishers.py` or `message_conversion.py` implement a singleton interface that expect parameters to be passed to them but they're either not passed or passed in a non-standard way.

This PR contains the following modifications:
- `Protocol` class now allows passing the parameters through the constructor instead of class variable
- All of the Capability classes retrieve their parameters through `protocol.parameters` upon initialization (No need to set `topics_glob` class variables) and use instance variables to get them later (e.g. `self.topics_glob` instead of `Advertise.topics_glob`).
- The default values are defined in class level and are only overwritten in the instance if `protocol.parameters` dictionary contains the keys.
- The ROS parameter and cli argument handling in RosbridgeWebsocketNode has been reworked:
  - Added new module constants SERVER_PARAMETERS and PROTOCOL_PARAMETERS that contain parameter name, type, default value and description.
  - Used `argparse` for parsing cli arguments (less boilerplate code and more user-friendly interface)
  - Made ROS parameters take precedence over cli args (passing cli args modifies the default values for ROS parameters)
  - A single class variable `protocol_parameters` is set in `RosbridgeWebSocket` handler, no need to create it on every opened connection
  - Added missing `actions_glob` parameter
- The `publishers.py` and `message_conversion.py` modules are now configured upon first initialization of the Protocol class.
 
 Some changes in behavior:
- `message_conversion.py` module needs to be configured before use. This is reflected in the `test_message_conversion.py` test.
- The `binary_encoder_type` and `bson_only_mode` are now actually passed to `message_conversion` module and have effect.
- Setting Node parameters before opening Protocol won't affect protocol parameters as the rosbridge_library does not use node_handle to retrieve parameters. All of parameters are passed in the dictionary in the Protocol constructor. This is reflected in tests.
- (Opinionated) `max_message_size` actually has effect now, that is, even when client sets the `fragment_size`, the actual fragment size is capped to `max_message_size` value.

~What's left to do:~
- ~I need to figure out how the bson parameters are handled, especially the relation between `binary_encoder` and `bson_only_mode`.~
- ~Some real life application testing~

~Even though this is still WIP, I would appreciate a review~


sea-bass:
I won't get to review this for a bit, but have you considered possibly using https://github.com/PickNikRobotics/generate_parameter_library?

Not sure how well this will do with hybrid ROS params and CLI arguments though, but at least may package up your params in a neat struct.

bjsowa:
> I won't get to review this for a bit, but have you considered possibly using https://github.com/PickNikRobotics/generate_parameter_library?
> 
> Not sure how well this will do with hybrid ROS params and CLI arguments though, but at least may package up your params in a neat struct.

Might be a good idea. I'll try to figure out whether I can use it together with cli args

bjsowa:
@sea-bass I looked at the possibility of using generate_parameter_library and IMO it's not worth it. It will only add complexity instead of simplifying it and won't provide any useful features. All of the parameters are read-only (read only once at the start) and don't require any validation. Also it's not possible to retrieve parameter descriptions for cli args before declaring ROS parameters.

bjsowa:
@sea-bass @EzraBrooks @MatthijsBurgh I think it's ready now. Would appreciate a review.

rosbridge_library/src/rosbridge_library/capability.py
```diff
@@ -68,6 +70,11 @@ def __init__(self, protocol: Protocol) -> None:
         """
         self.protocol = protocol
 
+        if self.parameter_names and self.protocol.parameters:
+            for param_name in self.parameter_names:
```
sea-bass: This and the next line could be a single list comprehension

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -112,41 +109,46 @@
...
+    :raises ValueError: If an unknown encoder type is specified.
+    """
+    global binary_encoder, binary_encoder_type, bson_only_mode, module_configured
+
+    if module_configured:
+        return
+
+    if parameters is not None:
+        if "binary_encoder" in parameters:
+            binary_encoder_type = parameters["binary_encoder"]
```
sea-bass: Rather than globals (yuck, though they were here before), I would do something like

```python
binary_encoder_type = parameters.get("binary_encoder", None)
```

(Of course the `, None` is not needed above, but putting there for other parameters that have different defaults

sea-bass: Although I see that those globals are used by the other free functions, so are they necessary? Hmmn, what if `configure()` simply returns the values it needs?

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -109,16 +114,27 @@ def __init__(self, client_id: str, node_handle: Node) -> None:
...
         self.operations: dict[str, Callable[[dict[str, Any]], None]] = {}
-        self.node_handle = node_handle
         self.external_service_list = {}
         self.external_action_list = {}
 
         if self.parameters:
-            self.fragment_size = self.parameters["max_message_size"]
-            self.delay_between_messages = self.parameters["delay_between_messages"]
-            self.bson_only_mode = self.parameters.get("bson_only_mode", False)
+            if "max_message_size" in self.parameters:
```
sea-bass: You can probably do repetitive stuff like this in a for loop over all the parameter names

rosbridge_library/test/internal/test_message_conversion.py
```diff
@@ -21,6 +21,10 @@
 
 
 class TestMessageConversion(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        c.configure()
```
sea-bass: Should this be `cls.configure()`?

Edit: oh no there is just a module being imported as `c` from before. Also gross and prone to misinterpretation, can we not alias that?

sea-bass (commented):
It does look cleaner! Have a few comments throughout.

Also for what it's worth, `generate_parameter_library` is not just read only once at the beginning. It can definitely make parameters runtime tunable. But it does rely on more tools and code generation that I am convinced are not necessary here.

rosbridge_library/src/rosbridge_library/capability.py
bjsowa: Could you show what specific you have in mind?

sea-bass:
```python
for param_name in self.parameter_names if param_name in self.protocol.parameters:
    setattr(self, param_name, self.protocol.parameters[param_name])

Comment (sea-bass) [reply rosbridge_library/src/rosbridge_library/capability.py]: ... actually I just noticed another option

```python
for param_name, param_value in self.parameter_names.items():
    if param_name in self.protocol.parameters:
        setattr(self, param_name, param_value)
```

this isn't a major comment, i was just staring at python

bjsowa: Is this actually a thing? I haven't seen such syntax in python

bjsowa: I just tested:
```
for name in list1 if name in list2:
    print(name) 
```
And it gave me syntax error

sea-bass: ugh no, I missed a thing. It would actually be this, which already makes it too hard to read

```python
for param_name in [p for p in self.parameter_names if p in self.protocol.parameters]:

File: rosbridge_library/src/rosbridge_library/internal/message_conversion.py (hunk 2 above)
Comment (bjsowa) [reply rosbridge_library/src/rosbridge_library/internal/message_conversion.py]: `binary_encoder_type` global is not used by any function so it can be removed

File: rosbridge_library/src/rosbridge_library/capability.py (hunk 1 above)
Comment (bjsowa) [reply rosbridge_library/src/rosbridge_library/capability.py]: Nah, this starts to look ugly. I might go with your other proposition though

Comment (sea-bass) [reply rosbridge_library/src/rosbridge_library/capability.py]: I think https://github.com/RobotWebTools/rosbridge_suite/pull/1060#discussion_r2404743781 may still be worthwhile though

File: rosbridge_library/test/internal/test_message_conversion.py (hunk 4 above)
Comment (bjsowa) [reply rosbridge_library/test/internal/test_message_conversion.py]: I agree the alias makes it less readable. I removed the aliases from the tests

File: rosbridge_library/src/rosbridge_library/protocol.py (hunk 3 above)
Comment (bjsowa) [reply rosbridge_library/src/rosbridge_library/protocol.py]: Something like that?
```Python
        if self.parameters:
            for param_name in ("max_message_size", "delay_between_messages", "bson_only_mode"):
                if param_name in self.parameters:
                    setattr(self, param_name, self.parameters[param_name])
```

sea-bass: exactly!

sea-bass (approved):

rosbridge_library/src/rosbridge_library/capability.py
bjsowa: > ... actually I just noticed another option
> 
> ```python
> for param_name, param_value in self.parameter_names.items():
>     if param_name in self.protocol.parameters:
>         setattr(self, param_name, param_value)
> ```
> 
> this isn't a major comment, i was just staring at python

I cannot use `items()` on `parameter_names` as it is a tuple of strings, not a dictionary. This would have to iterate over `parameters` instead:
```Python
        if self.parameter_names and self.protocol.parameters:
            for param_name, param_value in self.protocol.parameters.items():
                if param_name in self.parameter_names:
                    setattr(self, param_name, param_value)
```

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
bjsowa: I removed `binary_encoder_type`, one less global. `binary_encoder` and `bson_only_mode` are used by the functions though.

bjsowa: I had an idea for a refactor which would add a `MessageConverter` class that would have the `extract_values` and `populate_instance` methods and would take the protocol parameters in the constructor. The `Protocol` class would create an instance of the converter and use it instead of calling the free functions. This would allow to configure the converter with different parameters per Protocol instance (allowing for example using BSON and JSON at the same time). 

When I started refactoring it, it turned out it would require changes in many places and I decided to not do this to not overcomplicate this PR.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -112,41 +109,45 @@
...
-            node_handle.get_parameter_or("binary_encoder", Parameter("", value="default"))
-            .get_parameter_value()
-            .string_value
-        )
-        bson_only_mode = (
-            node_handle.get_parameter_or("bson_only_mode", Parameter("", value=False))
-            .get_parameter_value()
-            .bool_value
-        )
+    ;param parameters: A dictionary of parameters to configure the module.
```
MatthijsBurgh:
```suggestion
    :param parameters: A dictionary of parameters to configure the module.
```

rosbridge_library/src/rosbridge_library/protocol.py
```diff
@@ -222,8 +235,7 @@ def incoming(self, message_string: str = "") -> None:
         # This way, a client can change/overwrite its active values anytime by just including parameter field in any
         # message sent to rosbridge. Maybe need to be improved to bind parameter values to specific operation.
         if "fragment_size" in msg:
-            self.fragment_size = msg["fragment_size"]
-            # print "fragment size set to:", self.fragment_size
+            self.fragment_size = min(msg["fragment_size"], self.max_message_size)
```
MatthijsBurgh, line 238: What happens when a message has more data than the max message size?

rosbridge_library/src/rosbridge_library/rosbridge_protocol.py
```diff
@@ -83,7 +83,6 @@ class RosbridgeProtocol(Protocol):
     def __init__(
         self, client_id: str, node_handle: Node, parameters: dict[str, Any] | None = None
     ) -> None:
-        self.parameters = parameters
-        Protocol.__init__(self, client_id, node_handle)
+        Protocol.__init__(self, client_id, node_handle, parameters)
```
MatthijsBurgh: Why not use `super().__init__(...)`?

rosbridge_server/src/rosbridge_server/websocket_handler.py
```diff
@@ -181,8 +180,8 @@ def on_message(self, message: str | bytes) -> None:
 
     @log_exceptions
     def on_close(self) -> None:
-        assert self.node_handle is not None
         cls = self.__class__
+        assert cls.node_handle is not None
```
MatthijsBurgh: I assume the node handle should be a specific class. If so, better to check the node handle is an instance of that class.

rosbridge_server/src/rosbridge_server/websocket_handler.py
```diff
@@ -200,8 +199,8 @@ def send_message(self, message: bson.BSON | bytearray | str, compression: str =
         _io_loop.add_callback(partial(self.prewrite_message, message, binary))
 
     async def prewrite_message(self, message: bson.BSON | bytearray | str, binary: bool) -> None:
-        assert self.node_handle is not None
         cls = self.__class__
+        assert cls.node_handle is not None
```
MatthijsBurgh: Same as above

rosbridge_server/src/rosbridge_server/websocket_handler.py
```diff
@@ -124,38 +128,33 @@ def run(self) -> None:
...
-        assert self.node_handle is not None
         cls = self.__class__
-        parameters = {
-            "fragment_timeout": cls.fragment_timeout,
-            "delay_between_messages": cls.delay_between_messages,
-            "max_message_size": cls.max_message_size,
-            "unregister_timeout": cls.unregister_timeout,
-            "bson_only_mode": cls.bson_only_mode,
-        }
+        assert cls.node_handle is not None, "Node handle must be set before opening a WebSocket"
```
MatthijsBurgh: Same as below. Keep assertion messages the same between similar checks

MatthijsBurgh (changes requested):

rosbridge_library/src/rosbridge_library/protocol.py
bjsowa: It gets fragmented into multiple messages

rosbridge_server/src/rosbridge_server/websocket_handler.py
bjsowa: done

rosbridge_library/src/rosbridge_library/rosbridge_protocol.py
bjsowa: changed it

MatthijsBurgh (approved):

PR #980, opened 2024-12-17
Add ament_mypy test and fix all mypy errors

bjsowa:
**Public API Changes**
None

**Description**
Added ament_mypy test to CMakeLists for static type checking and fixed all of the errors reported by that checker.

PR #1063, opened 2025-08-13
[Humble] refactor: Enable ruff and ament_mypy checks and fix lint errors

bjsowa:
This is a backport of #1059, #980 and #1039 and an attempt to make `humble` branch more aligned with `ros2` branch.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,71 +7,73 @@ line-length = 100
 
 [tool.ruff.lint]
 select = [
-    # "AIR", # Airflow
-    # "ERA", # Eradicate
-    # "FAST", # FastAPI
+    # "AIR",  # Airflow, We don't use this framework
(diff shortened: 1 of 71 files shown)


sea-bass (approved):
I will keep feeling bad about not just making that import conditional, rather than cutting a separate Humble branch.

bjsowa:
> I will keep feeling bad about not just making that import conditional, rather than cutting a separate Humble branch.

Could you remind me what was the original reason for branching out?

sea-bass:
> Could you remind me what was the original reason for branching out?

It was https://github.com/RobotWebTools/rosbridge_suite/pull/932 I believe

bjsowa:
> > Could you remind me what was the original reason for branching out?
> 
> It was #932 I believe

I see. That's why I'm not a fan of using ros2cli internal APIs.

PR #1069, opened 2025-08-31
refactor: Add type annotations to all functions and methods

bjsowa:
**Public API Changes**
Aside from specifying type for public API arguments, the names of any arguments are still thead same.

**Description**
This was exhausting to do but IMO worth it. It always bothered me that I had to go back and forth between files to figure out what types does rosbridge implementation expect. 

**Changes**
- Added `from __future__ import annotations` to every source file - for better compatibility and flexibility in type hinting.
- Set `required-python` version to 3.10 - so ruff checks don't try to keep compatibility with earlier versions
- Enabled `flake8-annotations` check rules which enforce type hints for every argument and return types in all functions and methods.
- Fixed ruff check errors and new mypy errors

**Additional changes**
During the process of adding annotations and fixing type errors, I've made a few implementation changes that might affect the behavior of the library. Most notable changes are:

- In `rosbridge_library.capabilities.advertise.Registration` class,  added dummy advertise ID to use in place of None when the client does not provide id of the advertisement
- In `rosbridge_library.capabilities.defragmentation` module, got rid of `ReceivedFragments` singleton class and instead, added `lists` class variable to `Defragment` capability
- Simplified logging in  `rosbridge_library.capabilities.defragmentation.Defragment` class, making use of f-strings where applicable
- Multiple changes to `rosbridge_library.internal.message_conversion` module:
  - The `extract_values` function now calls `_from_object_inst` instead of `_from_inst` and `populate_instance` calls `_to_object_inst` instead of `_to_inst`. We can do this small optimization, assuming the `inst` passed to these function is a ROS message type. 
  - I added `_from_primitive_inst` to split the logic and make it more readable.
  - Improved `_to_time_inst` a little, raising errors when incorrect type name or message, creating Time and Duration message instances directly, instead of calling `rclpy.time.Time().to_msg()`
  - Added isinstance asserts to narrow down union types for static type checking. This can have a small but IMO negligible impact on performance, and can cause some false positives if I didn't predict that some types can be allowed at certain places.
  - Added helper `_remove_list_indicators` function

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -245,7 +245,7 @@ def _from_inst(
 
     # Check for primitive types
     if rostype in ros_primitive_types:
-        assert isinstance(inst, primitive_types | bytes), "Expected primitive type instance"
+        assert isinstance(inst, (*primitive_types, bytes)), "Expected primitive type instance"
         return _from_primitive_inst(inst, rostype)
 
     # Check if it's a list or tuple
@@ -310,7 +310,7 @@ def _from_object_inst(inst: ROSMessage, _rostype: str) -> dict:
 
 
 def _to_inst(
-    msg: dict[str, Any] | Sequence | str | float | bool,
+    msg: dict[str, Any] | Sequence | PrimitiveType,
     rostype: str,
     roottype: str,
     clock: Clock | None = None,
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -342,7 +342,7 @@ def _to_inst(
         return _to_primitive_inst(msg, rostype, roottype, stack)
 
     # Check whether we're dealing with a list type
-    if inst is not None and isinstance(inst, list | np.ndarray | array.array):
+    if inst is not None and isinstance(inst, list_types):
         assert isinstance(msg, list_types)
         return _to_list_inst(msg, rostype, roottype, clock, inst, stack)
 
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -151,7 +151,7 @@ def get_encoder() -> Callable[[ListType], bytes]:
 
 
 class InvalidMessageException(Exception):
-    def __init__(self, inst: object) -> None:
+    def __init__(self, inst: Any) -> None:  # noqa: ANN401
         Exception.__init__(
             self,
             f"Unable to extract message values from {type(inst).__name__} instance",
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -38,7 +38,7 @@ import re
 import sys
 from base64 import standard_b64decode, standard_b64encode
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any
 
 import numpy as np
 from builtin_interfaces.msg import Duration as DurationMsg
@@ -52,6 +52,8 @@ from rosbridge_library.internal.type_support import ROSMessage
 from rosbridge_library.util import bson
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from rclpy.node import Node
 
 
@@ -238,12 +240,12 @@ def _from_inst(
 
     # Check for time or duration
     if rostype in ros_time_types:
-        assert isinstance(inst, (TimeMsg, DurationMsg)), "Expected Time or Duration instance"
+        assert isinstance(inst, TimeMsg | DurationMsg), "Expected Time or Duration instance"
         return {"sec": inst.sec, "nanosec": inst.nanosec}
 
     # Check for primitive types
     if rostype in ros_primitive_types:
-        assert isinstance(inst, (primitive_types, bytes)), "Expected primitive type instance"
+        assert isinstance(inst, primitive_types | bytes), "Expected primitive type instance"
         return _from_primitive_inst(inst, rostype)
 
     # Check if it's a list or tuple
@@ -328,8 +330,8 @@ def _to_inst(
 
     # Check the type for time or rostime
     if rostype in ros_time_types:
-        assert isinstance(msg, (dict, str)), "Expected dict or str for time or duration msg"
-        assert inst is None or isinstance(inst, (TimeMsg, DurationMsg)), (
+        assert isinstance(msg, dict | str), "Expected dict or str for time or duration msg"
+        assert inst is None or isinstance(inst, TimeMsg | DurationMsg), (
             "Expected Time or Duration instance"
         )
         return _to_time_inst(msg, rostype, clock, inst)
@@ -340,7 +342,7 @@ def _to_inst(
         return _to_primitive_inst(msg, rostype, roottype, stack)
 
     # Check whether we're dealing with a list type
-    if inst is not None and isinstance(inst, (list, np.ndarray, array.array)):
+    if inst is not None and isinstance(inst, list | np.ndarray | array.array):
         assert isinstance(msg, list_types), "Expected a list msg"
         return _to_list_inst(msg, rostype, roottype, clock, inst, stack)
 
--- a/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/message_conversion.py
@@ -234,18 +234,18 @@ def _from_inst(
     # Special case for uint8[], we encode the string
     for binary_type, expression in ros_binary_types_list_braces:
         if expression.sub(binary_type, rostype) in ros_binary_types:
-            assert isinstance(inst, list_types), "Expected a list type instance"
+            assert isinstance(inst, list_types)
             encoded = get_encoder()(inst)
             return encoded.decode("ascii")
 
     # Check for time or duration
     if rostype in ros_time_types:
-        assert isinstance(inst, TimeMsg | DurationMsg), "Expected Time or Duration instance"
+        assert isinstance(inst, TimeMsg | DurationMsg)
         return {"sec": inst.sec, "nanosec": inst.nanosec}
 
     # Check for primitive types
     if rostype in ros_primitive_types:
-        assert isinstance(inst, (*primitive_types, bytes)), "Expected primitive type instance"
+        assert isinstance(inst, (*primitive_types, bytes))
         return _from_primitive_inst(inst, rostype)
 
     # Check if it's a list or tuple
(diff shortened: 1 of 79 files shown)


bjsowa:
@ros-pull-request-builder retest this please

bjsowa:
This fails in Jazzy, due to lack of support for type arguments in Future, Client, Subscription and other classes. @sea-bass Do you think it would be a good idea to branch out for `jazzy` now?

sea-bass:
> This fails in Jazzy, due to lack of support for type arguments in Future, Client, Subscription and other classes. @sea-bass Do you think it would be a good idea to branch out for `jazzy` now?

I'm OK with branching out!

sea-bass (approved):

bjsowa:
This is a massive PR, so I'll be waiting on some more reviews

MatthijsBurgh (approved):
I skimmed over it, I don't see any weird stuff. Anything I should pay special focus to? Otherwise I approve.

bjsowa:
> I skimmed over it, I don't see any weird stuff. Anything I should pay special focus to? Otherwise I approve.

Probably worth double checking message_conversion module.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -127,29 +144,32 @@ def configure(node_handle=None):
             sys.exit(0)
 
 
-def get_encoder():
+def get_encoder() -> Callable[[ListType], bytes]:
```
MatthijsBurgh, line 147: What is the bracket around `ListType` doing?

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -127,29 +144,32 @@ def configure(node_handle=None):
...
     configure()
+    assert binary_encoder is not None, "Binary encoder is not configured"
     return binary_encoder
 
 
 class InvalidMessageException(Exception):
-    def __init__(self, inst):
+    def __init__(self, inst: object) -> None:
```
MatthijsBurgh: In my opinion object doesn't have a place anymore in python since python 3. I think `Any` would be more suitable here.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -207,71 +229,78 @@ def msg_class_type_repr(msg_class):
...
     # Special case for uint8[], we encode the string
     for binary_type, expression in ros_binary_types_list_braces:
         if expression.sub(binary_type, rostype) in ros_binary_types:
+            assert isinstance(inst, list_types)
             encoded = get_encoder()(inst)
             return encoded.decode("ascii")
 
     # Check for time or duration
     if rostype in ros_time_types:
+        assert isinstance(inst, TimeMsg | DurationMsg)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -207,71 +229,78 @@ def msg_class_type_repr(msg_class):
...
-            and (math.isnan(inst) or math.isinf(inst))
-        ):
-            return None
-
-        # JSON does not support byte array. They are converted to int
-        if (not bson_only_mode) and (rostype == "octet"):
-            return int.from_bytes(inst, "little")
-
-        return inst
+        assert isinstance(inst, (*primitive_types, bytes))
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -207,71 +229,78 @@ def msg_class_type_repr(msg_class):
...
+        return _from_primitive_inst(inst, rostype)
 
     # Check if it's a list or tuple
-    if type(inst) in list_types:
+    if isinstance(inst, list_types):
         return _from_list_inst(inst, rostype)
 
     # Assume it's otherwise a full ros msg object
+    assert isinstance(inst, ROSMessage)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -207,71 +229,78 @@ def msg_class_type_repr(msg_class):
...
 
 
-def _from_list_inst(inst, rostype):
+def _from_primitive_inst(inst: PrimitiveType | bytes, rostype: str) -> PrimitiveType | bytes | None:
+    if bson_only_mode:
+        return inst
+
+    # JSON does not support Inf and NaN. They are mapped to None and encoded as null
+    if rostype in type_map["float"]:
+        assert isinstance(inst, float)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -207,71 +229,78 @@ def msg_class_type_repr(msg_class):
...
+        if math.isnan(inst) or math.isinf(inst):
+            return None
+
+    # octet is translated to byte array with length 1
+    # JSON does not support byte array. They are converted to int
+    if rostype == "octet":
+        assert isinstance(inst, bytes)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -290,28 +326,37 @@ def _to_inst(msg, rostype, roottype, clock=None, inst=None, stack=None):
...
         if expression.sub(binary_type, rostype) in ros_binary_types:
+            assert isinstance(msg, Sequence)
             return _to_binary_inst(msg)
 
     # Check the type for time or rostime
     if rostype in ros_time_types:
+        assert isinstance(msg, dict | str)
+        assert inst is None or isinstance(inst, TimeMsg | DurationMsg), (
+            "Expected Time or Duration instance"
+        )
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -290,28 +326,37 @@ def _to_inst(msg, rostype, roottype, clock=None, inst=None, stack=None):
...
         return _to_time_inst(msg, rostype, clock, inst)
 
     # Check to see whether this is a primitive type
     if rostype in ros_primitive_types:
+        assert isinstance(msg, primitive_types)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -290,28 +326,37 @@ def _to_inst(msg, rostype, roottype, clock=None, inst=None, stack=None):
...
         return _to_primitive_inst(msg, rostype, roottype, stack)
 
     # Check whether we're dealing with a list type
-    if inst is not None and type(inst) in list_types:
+    if inst is not None and isinstance(inst, list_types):
+        assert isinstance(msg, list_types)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -290,28 +326,37 @@ def _to_inst(msg, rostype, roottype, clock=None, inst=None, stack=None):
...
         return _to_list_inst(msg, rostype, roottype, clock, inst, stack)
 
     # Otherwise, the type has to be a full ros msg type, so msg must be a dict
     if inst is None:
         inst = ros_loader.get_message_instance(rostype)
 
+    assert isinstance(msg, dict)
+    assert isinstance(inst, ROSMessage)
```
MatthijsBurgh: I think asserts are not favorable, as they will just raise an `AssertionError` instead of manually raising a `TypeError`.

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
```diff
@@ -420,10 +489,41 @@ def _to_object_inst(msg, rostype, roottype, clock, inst, stack):
...
+        return list_match.group(1)
+
+    # Try bounded array pattern: type[length] or type[]
+    bounded_match = re.search(bounded_array_tokens, rostype)
+    if bounded_match:
+        return bounded_match.group(1)
+
+    # If neither pattern matches, this isn't a valid list type
+    err_msg = f"Invalid rostype for list: {rostype}"
+    raise ValueError(err_msg)
```
MatthijsBurgh, line 555:
```suggestion
    msg = f"Invalid rostype for list: {rostype}"
    raise ValueError(msg)
```

MatthijsBurgh (changes requested):

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
bjsowa: Callable takes 2 type arguments: a list of argument types and the return type. The brackets create a list with a single item

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
bjsowa: Due to [ANN401](https://docs.astral.sh/ruff/rules/any-type/) check, we can't use unconstrained type for function arguments

MatthijsBurgh: I would disable it in this line

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
bjsowa: `msg` variable name is used throughout this module for JSON messages, I used `err_msg` for error messages to differentiate between the two and make static type checker happy

rosbridge_library/src/rosbridge_library/internal/message_conversion.py
bjsowa: ok, done

bjsowa:
@MatthijsBurgh How do you propose to modify the type asserts so they raise TypeErrors?

MatthijsBurgh:
> @MatthijsBurgh How do you propose to modify the type asserts so they raise TypeErrors?

i.e.

```py
# assert isinstance(msg, dict)
if not isinstance(msg, dict):
    err_msg = f"msg is not a dict, but a {type(msg)}"
    raise TypeError(err_msg)
```

bjsowa:
@MatthijsBurgh Can you check it now and accept if it looks good to you?

MatthijsBurgh (approved):

Issue #1055, opened 2025-07-28
[DEVELOPERS ONLY] Who is working on which repository


MatthijsBurgh:
The teams in this organization are a bit of mess. I am trying to improve this. To get a better overview, please use this issue to let me know on which repositories you are working.

Please also let me know what level of permissions you think you should have. Or what level other people should (not) have.

I also hope to assign maintainers/admin for each repository.

Please share this issue to all members of this organization

bjsowa:
I maintain this repository and web_video_server.

EDIT: Since recently, also tf2_web_republisher

MatthijsBurgh:
@sea-bass @minggangw @gonzalocasas @chfritz @amacneil @EzraBrooks please provide insight on your work in this organization. Please forward this issue to anyone you are working with on any of our repos which I forgot to mention here.

chfritz:
@MatthijsBurgh: Of the repositories actually owned by this organization I'm mostly contributing to the web site (https://github.com/RobotWebTools/robotwebtools.github.io). Aside from this I'm the main author of Transitive and transAct, and contribute to rosnodejs, all of which we list on the web page.

sea-bass:
I'm not really using these tools at the moment, but am happy to keep helping out with basic maintenance tasks like PR review and unblocking CI. As such, I would like to still have some privileges to continue helping others get their work through. Thanks @MatthijsBurgh!

EzraBrooks:
We (MoveIt Pro team at PickNik) are using rosbridge, tf2_web_republisher, and roslibjs. We also recommend roslibpy to external integrators with our system, though don't use it directly ourselves.

I am happy to continue doing occasional PR reviews and we as a team will continue submitting fixes to bugs we find.

gonzalocasas:
I originally developed and currently maintain `roslibpy`, I have had a bit of a tight time, but I do want to continue supporting the maintenance of the repo

EzraBrooks:
At least for the moment, I'm less busy at my new role than my last one and am investing a bit more time here to clean up the roslibjs backlog - and if I can chew through that I might move onto some tf2_web_republisher and rosbridge stuff.

bjsowa:
What is there to do with `tf2_web_republisher` ?

EzraBrooks:
Ah, whoops, I meant ros3djs - I typed tf2_web_republisher because the most recent issue I saw regarding ros3djs was incompatibility with the ROS 2 version of the TF republisher :)

[...]

PR #1075, opened 2025-09-08
chore: Remove jazzy ci workflow

bjsowa:
Jazzy has now its separate branch

The failing rolling workflow is not related to this change. It will be fixed in #1069 which is the reason for the Jazzy split so this needs to be merged first

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,7 @@ jobs:
         include:
           # Test supported ROS 2 distributions
           # https://docs.ros.org/en/rolling/Releases.html
+          # NOTE: Humble and Jazzy do not work on the `ros2` branch, so they are tested in their own branches.
           - ros: kilted
             os: ubuntu-24.04
           - ros: rolling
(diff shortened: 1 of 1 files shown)


bjsowa:
@EzraBrooks @MatthijsBurgh Is there an option I could get admin access to this repo? This would help me do some maintenance stuff related to workflows.

.github/workflows/ci.yml
```diff
@@ -21,9 +21,6 @@ jobs:
         include:
           # Test supported ROS 2 distributions
           # https://docs.ros.org/en/rolling/Releases.html
-          # NOTE: Humble does not work on the `ros2` branch, so it is tested in its own branch.
```
sea-bass, line 24: I think you want to keep this comment, but amend it with "Jazzy and Humble" right?

bjsowa: I removed to not have to update it each time we branch out. But you're right, should probably leave this in.

sea-bass (approved):

MatthijsBurgh (approved):

MatthijsBurgh:
@bjsowa I will update permissions soon based on #1055 
I have updated the branch protection for the ros2 and jazzy branch

bjsowa:
> @bjsowa I will update permissions soon based on #1055 I have updated the branch protection for the ros2 and jazzy branch

Ok, can you merge this PR bypassing the rules for now?

PR #1098, opened 2025-11-05
[Humble] feat: Add type annotations, new parameter handling (backport #1069, #1060)

bjsowa:
This is a backport of #1069 and #1060. It was easier to bundle them together to avoid having to fix some mypy errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[project]
+name = "rosbridge_suite"
+requires-python = ">=3.10"
+
 [tool.black]
 # We use ruff for formatting, but black formatter should be compatible with ruff.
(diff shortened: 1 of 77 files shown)


EzraBrooks (commented):
For things like this, what CI job should I be looking at? I see a ruff check, but am I missing a type checker like ty or pyright here?

bjsowa:
> For things like this, what CI job should I be looking at? I see a ruff check, but am I missing a type checker like ty or pyright here?

We use ament_mypy which is run as part of cmake tests. (CI runs `colcon test`)

EzraBrooks (approved):
Ah, I'd forgotten about ament_lint and its ilk.

PR #1107, opened 2025-11-13
feat: Replace cbor with cbor2 library, remove inline implementation

### LINKED ISSUE #1106: Switch to cbor2 python package and drop the inline implementation of CBOR
Inline CBOR implementation was added at one point due to the cbor package not being available on fedora (#377). The cbor2 package is slightly faster and is available on all major distros.

bjsowa:
**Public API Changes**
None

**Description**
Closes #1106


EzraBrooks (approved):
Sweet!

sea-bass:
amazing!

PR #1108, opened 2025-11-14
feat: Allow importing action feedback message types

### LINKED ISSUE #1072: rosbridge can’t subscribe to ROS 2 action topic  `/navigate_to_pose/_action/feedback`
**Description**

* **Library Version:** rosbridge suite v2.3.0 (binaries installed using apt: `ros-rolling-rosbridge-msgs ros-rolling-rosbridge-library ros-rolling-rosbridge-server ros-rolling-rosbridge-suite`), roslibjs npm 1.4.1 (we understand this version doesn't support Actions and only the `v2` version does, but the issue wasn't with roslib, added only for context), Nav2 build from src (Rolling)
* **ROS Version:** ROS 2 Rolling (Debian)
* **Platform / OS:** Ubuntu 24.04

Subscribing to the ROS 2 action feedback topic `/navigate_to_pose/_action/feedback` (type `nav2_msgs/action/NavigateToPose_FeedbackMessage`) from rosbridge throws the error

```
[rosbridge_websocket-1] [ERROR] [1756883713.567298692] [rosbridge_websocket]: [Client fc2f145c-2bf9-47e4-82dc-1c08d08975e5] [id: subscribe:/navigate_to_pose/_action/feedback:22] subscribe: Unable to import msg class NavigateToPose_FeedbackMessage from package nav2_msgs. Caused by module 'nav2_msgs.msg' has no attribute 'NavigateToPose_FeedbackMessage'
```

since rosbridge tries to import from `nav2_msgs.msg` instead of `nav2_msgs.action`.

This seems like rosbridge’s type loader is looking for `nav2_msgs.msg.NavigateToPose_FeedbackMessage`, but in ROS 2 these wrapper types used by these actions are under `nav2_msgs.action`.

> RViz nav2 panel (here [https://github.com/ros-navigation/navigation2/blob/main/nav2\_rviz\_plugins/src/nav2\_panel.cpp](https://github.com/ros-navigation/navigation2/blob/main/nav2_rviz_plugins/src/nav2_panel.cpp)) does a similar thing. rclcpp subscribes to the action’s feedback topic using the generated C++ type `nav2_msgs::action::NavigateToPose::Impl::FeedbackMessage`.

> Rosbridge Action **API** works fine in our testing if the goal is sent via rosbridge. Meaning, if we don't use our pkg to send action goals. The problem here is **subscribe** to the hidden wrapper topic when goals are sent by another node.

@padhupradheep and I have been working on a GUI tool that uses WebSockets for communicating with ROS 2. Hence we decided to use rosbridge suite with roslibjs .We trigger navigation via our package neo\_waypoint\_follower by calling the /start\_waypoint\_loop service (std\_srvs/srv/Trigger, implemented by the waypoint\_looper node);

We are not sending the action goal via rosbridge, so rosbridge’s Action API callbacks (which are tied to a client-side goal handle) cannot be used in our setup. Instead, we try to subscribe directly to /navigate\_to\_pose/\_action/feedback, and hence the issue.

As a workaround, we could have the neo\_waypoint\_follower pkg node republish a feedback topics for the specific fields our GUI needs -estimated\_time\_remaining, distance\_remaining and navigation\_time and have the GUI subscribe to that. However, this is redundant because Nav2 already publishes these via the `/navigate_to_pose/_action/feedback` topic.

Looking at `rosbridge_library/internal/ros_loader.py`, the message loader for subscribe paths might be importing from `pkg.msg`. For types provided as `pkg/action/Name_*` it could instead import from `pkg.action`.

This would bring subscribing capability close to what was done in RViz’s Nav2 panel that subscribes to `nav2_msgs::action::NavigateToPose::Impl::FeedbackMessage`.

Please advise on the correct path forward. Is this a rosbridge limitation or bug in the subscribe path for ROS 2 action wrapper message types, or are we misusing the API?
Happy to provide additional logs or context. Thanks in advance.

---

**Steps To Reproduce**

1. With a navigate\_to\_pose task active, we verified the hidden feedback topics exist:

   ```bash
   ros2 topic list -a | grep navigate_to_pose
   # includes  /navigate_to_pose/_action/feedback
   ros2 topic info -t /navigate_to_pose/_action/feedback
   # Type: nav2_msgs/action/NavigateToPose_FeedbackMessage
   ```

2. Connecting a roslibjs client and subscribe:

   ```js
   new ROSLIB.Topic({
     ros,
     name: '/navigate_to_pose/_action/feedback',
     messageType: 'nav2_msgs/action/NavigateToPose_FeedbackMessage',
   }).subscribe(msg => {
     const fb = msg.feedback || msg;
     const secs = t => (t?.sec ?? 0) + (t?.nanosec ?? 0) / 1e9;
     // consume fb.navigation_time, fb.estimated_time_remaining, fb.distance_remaining …
   });
   ```

   This is what causes the following error.

3. In our launch file we do:

   ```py
   IncludeLaunchDescription(
           XMLLaunchDescriptionSource(
               os.path.join(get_package_share_directory('rosbridge_server'), 'launch', 'rosbridge_websocket_launch.xml')
               )
           )
   ```

---

**Expected Behavior**

Rosbridge should successfully import the wrapper type from the `*.action` module and deliver messages to the client, identical to native rclcpp/rclpy subscribers and RViz.

---

**Actual Behavior**

the error message

```
[rosbridge_websocket-1] [ERROR] [1756883713.567298692] [rosbridge_websocket]: [Client fc2f145c-2bf9-47e4-82dc-1c08d08975e5] [id: subscribe:/navigate_to_pose/_action/feedback:22] subscribe: Unable to import msg class NavigateToPose_FeedbackMessage from package nav2_msgs. Caused by module 'nav2_msgs.msg' has no attribute 'NavigateToPose_FeedbackMessage'
```

Type import fails because rosbridge attempts to import from the `*.msg` module:

```
nav2_msgs.msg.NavigateToPose_FeedbackMessage  # (does not exist)
```

bjsowa:
**Public API Changes**
None

**Description**
Fixes #1072 

Actions in ROS2 use hidden topics for publishing feedback messages. The feedback message classes are in the form `example_interfaces.action.Fibonacci.Impl.FeedbackMessage` which suggests they are not the part of public API but ROS2 cli tools support introspecting these messages and they denote the message type as `example_interfaces/action/Fibonacci_FeedbackMessage` which is an exception to the rule which says every message type has `msg` in the middle part.

This PR adds support for using action feedback message types for topics in rosbridge. It does this by treating this as a special case, similarly how it is [treated in rosidl_runtime_py](https://github.com/ros2/rosidl_runtime_py/blob/rolling/rosidl_runtime_py/import_message.py#L34-L36)


sea-bass (approved):

EzraBrooks (approved):
Seems technically correct, though I worry that it's encouraging usage of a private API.. but I guess other tools already allow that.

PR #1122, opened 2025-11-18
Document Base64 Encoding

### LINKED ISSUE #1121: Documentation Does Not Indicate Base64 String Encoding for uint8[]
**Description**

Protocol documentation fails to mention base64 encoding for byte arrays.

Got a bug report for roslibrust that image messages were failing to pass through correctly:

https://github.com/RosLibRust/roslibrust/issues/278

This turned out to be because of an unfortunate interaction with deserializer we were using and base64 encoded strings, where it did successfully decode to a Vec<u8>, but the data was mangled because it wasn't actually decoded.

Would be willing to update the protocol specification to include information about base64 encoding of byte arrays, but wanted to confirm that this is the intended behavior.

- Library Version: roslibrust - ALL
- ROS Version: Confirmed issue present on Noetic, Iron, Kilted (assume all)
- Platform / OS: Ubuntu 24.04

**Steps To Reproduce**

See linked roslibrust issue, server appears to accept a JSON array of numbers for a uint8[] message field, but will return a base64 encoded string.

**Expected Behavior**

Behavior here is fine, just thrown for a loop that it isn't clearly documented.

**Actual Behavior**

Carter12s:
**Public API Changes**

None.

**Description**

Documentation update.

#1121

### DIFF
--- a/ROSBRIDGE_PROTOCOL.md
+++ b/ROSBRIDGE_PROTOCOL.md
@@ -613,3 +613,25 @@ The meta-package will contain the following packages:
     tornado, a python server implementation.
  * **rosapi** – provides ROS services for various master API calls, such as
     listing all the topics, services, types currently in ROS
+
+### 4.6 Data Type Encoding
+
+**Base64 Encoding for JSON Byte Arrays**
+
+When rosbridge sends messages containing `uint8[]` or `char[]` fields as JSON, these byte arrays are automatically encoded as base64 strings to reduce message size by approximately 60-65%.
+
+For example, a message field defined as:
+
+```
+uint8[] data = [0, 0, 0, 0]
+```
+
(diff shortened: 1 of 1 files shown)


EzraBrooks (approved):
LGTM!

bjsowa:
@EzraBrooks For future reference, I try to use semantic tags for commit messages (`feat:`, `fix:`, `chore:`, `docs:` etc.). Was pretty consistent the last few month.

EzraBrooks:
Sounds like we should add a CI job. https://github.com/commit-check/commit-check

bjsowa:
> Sounds like we should add a CI job. https://github.com/commit-check/commit-check

I wouldn't want to enforce naming convention for individual commits in the PR since we squash merge them either way.

PR #1133, opened 2025-11-20
fix: Invert success check for set_param function

bjsowa:
**Public API Changes**
None

**Description**
The `successful` field in `set_param` service was set to `false` when setting the parameter was successful. This bug was introduced in #1069


sea-bass (approved):
oops!

PR #1143, opened 2026-01-27
chore: disable colcon-powershell

bjsowa:
This is a hotfix for the CI failing due to `ModuleNotFoundError: No module named 'ament_package'` error (see [here](
https://github.com/RobotWebTools/rosbridge_suite/actions/runs/21110984875/job/60709468192#step:4:5589) for example). The workflow has been failing for about a month now and I suspect it is due to some recent change to the Github runner images. I couldn't find any other solution than running the job in a container.

The other test is failing due to recent changes in rclpy that changed some executor behavior (which I authored :sweat_smile:). It will be fixed in another PR later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
     name: ROS 2 ${{ matrix.ros }} (${{ matrix.container }})
     runs-on: ubuntu-latest
-    container: ${{ matrix.container }}
+    # container: ${{ matrix.container }}
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,7 @@ jobs:
 
     name: ROS 2 ${{ matrix.ros }} (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
+    container: ubuntu:noble
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
     name: ROS 2 ${{ matrix.ros }} (${{ matrix.container }})
     runs-on: ubuntu-24.04
-    # container: ${{ matrix.container }}
+    container: ${{ matrix.container }}
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
       - uses: ros-tooling/setup-ros@v0.7
 
       - uses: ros-tooling/action-ros-ci@v0.4
+        id: action_ros_ci_step
         with:
           target-ros2-distro: ${{ matrix.ros }}
 
@@ -69,3 +70,10 @@ jobs:
           which cmake
           echo -e "\e[35;1mcmake --version\e[0m"
           cmake --version
+
+      - uses: actions/upload-artifact@v6
+        with:
+          name: colcon-logs
+          path: ${{ steps.action_ros_ci_step.outputs.ros-workspace-directory-name }}/log
+        if: always()
+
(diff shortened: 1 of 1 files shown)


MatthijsBurgh:
In my mind it is very unlikely an update in a runner image causes the CI to break. But here is the diff between the image releases, [link](https://github.com/actions/runner-images/compare/ubuntu24/20251126.144...ubuntu24/20251215.174)

You can check it to see whether you see something interesting.

bjsowa:
> In my mind it is very unlikely an update in a runner image causes the CI to break. But here is the diff between the image releases, [link](https://github.com/actions/runner-images/compare/ubuntu24/20251126.144...ubuntu24/20251215.174)
> 
> You can check it to see whether you see something interesting.

I tried looking for something interesting there but couldn't find anything. I'm suspicious of runner image for two reasons:
- Running the CI in `ubuntu:noble` container image or using `ubuntu-slim` runner (see [here](https://github.com/RobotWebTools/rosbridge_suite/actions/runs/21445715995/job/61760800688)) does not result in the error.
- I have a private repo where the CI builds the whole ROS (including `ament_package`) in a Nix devshell which freezes all the dependencies and it started suffering from the same error around similar time.

MatthijsBurgh:
Can we print the filesystem tree or similar just before building? I am just looking for a way to debug this issue.

bjsowa:
@MatthijsBurgh Github actions seem to be down for now

MatthijsBurgh:
Yes Indeed 😢

bjsowa:
@MatthijsBurgh I have no more ideas for things I can check for now.

sea-bass:
honestly? I had CI fail on one of my repos and deleting all the github actions caches and rerunning solved it for me. Try that?

EDIT: oh, there are no caches here...

bjsowa:
> honestly? I had CI fail on one of my repos and deleting all the github actions caches and rerunning solved it for me. Try that?
> 
> EDIT: oh, there are no caches here...

There is only cache from pre-commit job. It shouldn't affect anything but I cleared it anyway and it did not change anything.

bjsowa:
@MatthijsBurgh Any idea how to debug this further?

MatthijsBurgh:
Preferably we can print the state midway through the ros-ci-action. Are there any hooks we can define?

bjsowa:
> Preferably we can print the state midway through the ros-ci-action. Are there any hooks we can define?

We could just try running the commands manually, instead of using the action

MatthijsBurgh:
We can try checking the colcon logs.

See [here](https://github.com/ros-tooling/action-ros-ci?tab=readme-ov-file#store-colcon-logs-as-build-artifacts) how to store the logs as artifacts

MatthijsBurgh:
@bjsowa when running on the runner the logs show the following PYTHONPATH

```
PYTHONPATH=/home/runner/work/rosbridge_suite/rosbridge_suite/ros_ws/install/rosbridge_test_msgs\lib/python3.12/site-packages;${PYTHONPATH}
```

The backslash and the semi-colon show that some part of the tooling, either colcon/ament or the action-ros-ci, thinks we are on Windows. I think this is something we should investigate.

MatthijsBurgh:
@bjsowa I don't know what changed in december causing the issues. But the issue is `colcon-powershell` not being skipped. It wasn't skipped, because the `PSModulePath` is set in the runner. [See the check](https://github.com/colcon/colcon-powershell/blob/1709d7d2e0fe856a038bff76f6159f169d891c03/colcon_powershell/shell/powershell.py#L69) For me it is hard to say whether that check is okay or too much of a shortcut.

But by overriding it in the CI config, we make it work for now.

Might be fixed by https://github.com/colcon/colcon-powershell/pull/42

MatthijsBurgh (dismissed):

bjsowa:
Thank you @MatthijsBurgh for finding the culprit! Should we remove the other changes or leave it as it is?

MatthijsBurgh:
I kept these changes as I think both changes (bash as default shell; upload log artifacts) are useful. So you can merge if you agree.

MatthijsBurgh (approved):

PR #1147, opened 2026-02-11
fix: Failing service and subscription tests

bjsowa:
**Public API Changes**
None

**Description**
Recent changes related to coroutines in rclpy (ros2/rclpy#1469) fixed (at least partially) exception propagation. Executor no longer throws an exception when the service handler awaits a future that resulted in exception (this is a good thing). This, however, uncovered another problem with the test as it falsy expects to receive a message after sending `unadvertise_service`.

This PR fixes the test and adds a comment explaining current limitations of the implementation.

EDIT: Also modified the subscription asserts to add a timeout. This should hopefully fix the flaky tests.


bjsowa:
I've rerun the workflow 10 times, no more randomly failing tests... I hope.

bjsowa:
@MatthijsBurgh Could you approve the changes?

MatthijsBurgh (approved):

Issue #887, opened 2023-11-07
Adding QoS to Publisher


yoelbassin:
Working over wi-fi can be challenging due to communication over a lossy or unstable network. ROS QoS is a great mechanism to handle these issues, with it's reliability, durability and lifespan option it can solve and help in many ways. An example is a topic with a reliable publisher with volatile durability. Another option is using BEST_EFFORT reliability to lower traffic throughput.

A base for this approach was already merged In #86, but it still lacks many features (the attributes I mentioned above). A possible solution is adding `durability`, `reliability` and `lifespan` attributes to the query, such that the new JSON will be

```json
{ "op": "publish",
  (optional) "id": <string>,
  "topic": <string>,
  (optional) "reliability": <string>,
  (optional) "durability": <string>,
  (optional) "lifespan": <string>,
  "msg": <json>,
}
```

or, a way better approach will be to send the entire QoS as an inner json in the message:

```json
{ "op": "publish",
  (optional) "id": <string>,
  "topic": <string>,
  (optional) "qos": <json>,
  "msg": <json>,
}
```
Where the `qos` JSON is
```json
{
  (optional) "reliability": <string>,
  (optional) "durability": <string>,
  (optional) "lifespan": <string>,
  ...
}
```

danaiger:
Seem like a good idea. It caught me as well when I had to create a publisher with custom life span message. Anyone has an idea if this is already implemented somehow?

rejero:
Following this, im running into issues where i need published messages sent via rosbrodge to have the same custom qos settings as my subscribers on ros2 nodes. Might need the same for my subscribers receiving via rosbridge too

yoelbassin:
@sea-bass what do you think about these options? I want to start working on a PR for this issue but it will change the API and would require changes in multiple repos as @JayHerpin mentioned in #908

The first solution (of just adding some options to the JSON) is fine and wouldn't create breaking changes, but it would require changing the API once again if more QoS options would be added. 

The second option of adding a QoS sub-JSON would just add a single parameter which could easily be extended later, but since there is already an option for `queue_size` (merged in #86) it would either require a breaking change, or a duplicate parameter that can be marked as deprecated.

Would love to hear opinion about this issue and the proposed options

sea-bass:
I would as much as possible keep the QoS configuration at the time of creating the publisher/subscriber itself, and not every time you send a single message payload. This mimics how ROS does it anyhow.

[...]

Issue #1123, opened 2025-11-18
Provide Mechanism to Access Protocol Version / API Version


Carter12s:
I'm maintaining `roslibrust` with a major goal being maintaining ROS1 and ROS2 compatibility.

Happy for the API to continue to evolve, but it would be really handy to have some mechanism to discover what rosbridge_server variant I'm talking with, and adapt my protocol appropriately.

Providing this a ros service, "/rosbridge_server/get_version" that returns the version from the package.xml would be a simple way to do this. But we may prefer "/rosbridge_server/protocol_version" that returns a SemVer number corresponding to the design of the ROSBRIDGE_PROTOCOL.md

Versioning the protocol would be useful for the documentation as well!

EzraBrooks:
Personally I'd propose hosting this as an HTTP GET endpoint on the same HTTP socket as the bridge hosts its WebSocket endpoint on.

I'd also like to see similar for "is the bridge in BSON mode" if we continue supporting BSON. Right now clients are blind as to whether they should be speaking BSON for every transaction or not - and if they try to speak JSON to a BSON-only bridge everything just explodes

Carter12s:
Yeah would make sense to have this operate as standalone super simple protocol, so that you don't have to have the plumbing for making a service_request working ahead of this call. A simple http GET, that returns a little json payload would be perfect.

```json
{
  server_version: "0.2.0", // Could do 3 numbers instead of string?
  api_version: "1.2.0", 
  supports_json: true,
  supports_cbor: true,
  supports_bson: false,
  supports_tls: false
}
```

bjsowa:
I agree with the premise. This could be particularly useful for #1150 to let the client know whether they can use the qos options.

> I'd also like to see similar for "is the bridge in BSON mode" if we continue supporting BSON. Right now clients are blind as to whether they should be speaking BSON for every transaction or not - and if they try to speak JSON to a BSON-only bridge everything just explodes

BSON was never documented so I treated it as not a part of the "official" protocol, hence I got rid of it in Rolling.

I'll try freeze the current protocol to version v2.0 and any change will result in an increase in the minor version (e.g. v2.1, v2.2), with the quick summary of changes between the versions at the top of the document.

> Personally I'd propose hosting this as an HTTP GET endpoint on the same HTTP socket as the bridge hosts its WebSocket endpoint on.

[...]

Issue #1012, opened 2025-02-20
Inconsistency between protocol doc and code


ahoenerBE:
**Description**
I'm currently working on a PR to RosLibJS and noticed some inconsistency between the protocol doc and the actual code.

e.g. In Advertising, rosbridge / roslib add fields `latch` and `queue_size` to the packet and are pulling them from the message

```py
# rosbridge_suite/rosbridge_library/src/rosbridge_library/capabilities/advertise.py
def advertise(self, message):
        # Pull out the ID
        aid = message.get("id", None)

        self.basic_type_check(message, self.advertise_msg_fields)
        topic = message["topic"]
        msg_type = message["type"]
        latch = message.get("latch", False)
        queue_size = message.get("queue_size", 100)
        # snip...
```

But the protocol document doesn't include these fields. Latch is mentioned briefly in its own section, and the phrase `queue_size` doesn't appear in the entire document.
```md
#### 3.3.1 Advertise ( _advertise_ )

If you wish to advertise that you are or will be publishing a topic, then use the advertise command.


{ "op": "advertise",
  (optional) "id": <string>,
  "topic": <string>,
  "type": <string>
}


 * **topic** – the string name of the topic to advertise
 * **type** – the string type to advertise for the topic

   * If the topic does not already exist, and the type specified is a valid
     type, then the topic will be established with this type.
   * If the topic already exists with a different type, an error status message
     is sent and this message is dropped.
   * If the topic already exists with the same type, the sender of this message
     is registered as another publisher.
   * If the topic doesn't already exist but the type cannot be resolved, then
     an error status message is sent and this message is dropped.
```

It seems like publishers/advertisement use queue_size and subscriptions use queue_length?

Is this intentional? Are these fields no longer supported? Looking for some answers to guide my development.

ahoenerBE:
@bjsowa @sea-bass sorry to bother with the ping, but I'd like to try and knock some of this out before the weekend if you have any idea if these extra arguments that roslibjs is passing along are still supported.  
Is it some sort of ROS1 compat?

bjsowa:
Sorry for a late reply. The `queue_size` and `latch` arguments for publishers are supported. In ROS1, these arguments are just passed to Publishers. In ROS2, they are translated to QoS policies (the names are left for backward compatibility):

- All publishers use trancient local durability and a lifespan of 1 second by default.
- `queue_size` is directly translated to QoS Depth policy.
[...]

PR #1150, opened 2026-02-18
feat: Add QoS Profile support for advertise, publish and subscribe operations

Roald-Schaum:
**Public API Changes**

register_advertisement now has a QoSProfile variable named 'qos'
subscribe now has a QoSProfile variable named 'qos'

If the qos variable is not supplied at any point in the pipeline, it defaults to the currently preexisting QOS settings.  

**Description**

Message topics now have support for QOS profiles. This was not a thing in ROS1, but with ROS 2 it is necessary to have compliant QOS profiles between all publishers and subscribers on a node.

https://github.com/RobotWebTools/rosbridge_suite/issues/887


bjsowa:
Thank you for the PR! QoS support has been a long awaited feature in Rosbridge. As this is extending the protocol, it will need more thorough discussion on the changes. I summon @EzraBrooks @MatthijsBurgh @sea-bass .

In the meantime, could you update the Protocol Specification file ([ROSBRIDGE_PROTOCOL.md](https://github.com/RobotWebTools/rosbridge_suite/blob/ros2/ROSBRIDGE_PROTOCOL.md)) ?

Roald-Schaum:
I sincerely apologize for how many commits this has been, I can't actually test any of the code on my own system because of a very abnormal ROS setup that I have, so I've been relying on the github actions to verify if my code actually works.

Roald-Schaum:
These tests failures are from the tests having incompatible QoS settings between the publishers and subscribers. They all pass if the change is made, but wasn't sure if it is my place to change them.

MatthijsBurgh:
I have seen this PR. I will try to review as soon as possible. But probably require 1 week.

rosbridge_library/src/rosbridge_library/internal/subscribers.py
```diff
@@ -207,21 +219,23 @@ def subscribe(
...
-                self.qos.reliability = ReliabilityPolicy.BEST_EFFORT
+            # Shouldn't need to attempt this now
+            # infos = self.node_handle.get_publishers_info_by_topic(self.topic)
+
+            # if len(infos) > 0 and all(
+            #     pub.qos_profile.durability == DurabilityPolicy.TRANSIENT_LOCAL for pub in infos
+            # ):
+            #     self.qos.durability = DurabilityPolicy.TRANSIENT_LOCAL
+            # if any(pub.qos_profile.reliability == ReliabilityPolicy.BEST_EFFORT for pub in infos):
+            #     self.qos.reliability = ReliabilityPolicy.BEST_EFFORT
```
EzraBrooks: remove commented code

ROSBRIDGE_PROTOCOL.md
```diff
@@ -370,24 +372,14 @@ which to send messages.
...
-being sent. Messages are sent from the head of the queue. If the queue gets
-full, the oldest message is removed and replaced by the newest message.
-
-If a client has multiple subscriptions to the same topic, then messages are
-sent at the lowest throttle_rate, with the lowest fragmentation size, and
-highest queue_length. It is recommended that the client provides IDs for its
-subscriptions, to enable rosbridge to effectively choose the appropriate
-fragmentation size and publishing rate.
+* **qos** - the qos profile of the topic to subscribe to. If not specified,
+    queue_depth is set to 10
```
EzraBrooks: If we're already changing the default queue depth from `100` to `10`, I would argue we should break the convention entirely and use a `QoSProfile` object (defaulting to the default QoS) for this state internally instead of that strange `QoSProfile | int` type.

Roald-Schaum: I agree that it's strange, but that's what the official [rclpy docs](https://docs.ros2.org/latest/api/rclpy/api/node.html#rclpy.node.Node.create_subscription) use.

The reason I changed the default queue_depth to 10 is because of the QoS wiki entry [kilted](https://docs.ros.org/en/kilted/Concepts/Intermediate/About-Quality-of-Service-Settings.html) (also on [rolling](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html)) states that is the default queue_depth. I'm open to changing those back, and removing the QoSProfile | int type, but I figured I'd try to make it more consistent with the more easily accessed docs.

ROSBRIDGE_PROTOCOL.md
```diff
@@ -352,9 +354,9 @@ then the current time will be automatically inserted.
   "topic": <string>,
   (optional) "type": <string>,
   (optional) "throttle_rate": <int>,
-  (optional) "queue_length": <int>,
   (optional) "fragment_size": <int>,
-  (optional) "compression": <string>
+  (optional) "compression": <string>,
+  (optional) "qos": <QoSProfile>
```
bjsowa: This doesn't really correspond to how the qos is retrieved in the implementation, where you try to resolve the `qos_depth`, `durability_policy`, etc.

Though, IMO it would look better as a nested structure, for example `qos.depth`, `qos.durability`, etc.

ROSBRIDGE_PROTOCOL.md
EzraBrooks: I'm more concerned about the strange union type than I am about the change in the default value, to be honest

Roald-Schaum:
Shall I modify the tests to have compatible QOS profiles between the publishers and subscribers?

bjsowa:
You are changing the default behavior of rosbridge which tries to adjust QoS for best compatibility. I would like to leave it as it is when custom qos is not provided.

For the protocol, I propose to use a nested JSON object for the qos profile. That way, it's easier to reuse the parsing implementation for different capabilities and we don't pollute the protocol specification that much. Moreover, I would prefer using a string representation of the policies instead of the raw enum numbers and we could add `lifespan` to the specification as well.

Here's an example of the advertise message in my proposal:
```JSON
{
  "op": "advertise",
  "topic": "/foo",
  "type": "std_msgs/msg/String",
  "qos": {
    "depth": 10,
    "reliability": "reliable",
    "durability": "volatile",
    "history": "keep_last",
    "lifespan": 1.5
  }
}
```
Not setting `qos` field should result in current default behavior of rosbridge. If `qos` is set but not all options are present, they should fallback to system default. In particular, setting `qos` to empty object (`qos: {}`) should result in system default qos profile.

@EzraBrooks @MatthijsBurgh @sea-bass I wonder what you guys think.

bjsowa:
One other thing to consider: What if two clients subscribe to the same topic but with different qos? Should the subscriber be overwritten or should there be one instance of `MultiSubscriber` per requested QoS?

Roald-Schaum:
> One other thing to consider: What if two clients subscribe to the same topic but with different qos? Should the subscriber be overwritten or should there be one instance of `MultiSubscriber` per requested QoS?

This seems to have been an issue discussed [previously](https://github.com/RobotWebTools/rosbridge_suite/pull/582) https://github.com/RobotWebTools/rosbridge_suite/pull/582. The already-documented behavior of ROS would be to have one instance per QoS profile as [this](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Quality-of-Service-Settings.html#qos-compatibilities) would ascribe. While it _is_ possible to overwrite QoS settings, I think that would be a much worse solution as it would only lead to some subscribers having to be dropped if new subscribers with different QoS settings came in, which I think should instead be left up to the user.

bjsowa:
@Roald-Schaum  I'm working on rewriting the protocol specification document to make it easier to read and fix some inconsistencies (see #1012) and providing a way to the client for access protocol version (see #1123).

Once that is done, we could proceed with this PR and make it increase the protocol version to v2.1

Roald-Schaum:
@bjsowa Sounds good. If you need me to do anything just let me know, I'm happy to do it.

bjsowa:
@Roald-Schaum I pushed my proposal for the protocol in SHC-ASTRA/astra_rosbridge_suite#3
Please review it and, if you don't have any objection, merge it. Then adjust your implementation to follow the specification.

bjsowa:
@Roald-Schaum My review would require a long back-and-forth so to push this feature faster I decided to develop my changes on top of your branch. Please check out SHC-ASTRA/astra_rosbridge_suite#5 and if you don't have any objections, merge it to your branch.

bjsowa (approved):

bjsowa:
LGTM obviously. Since this PR adds a big feature to the protocol, I'm curious about what authors of the client libraries have to say: @EzraBrooks @gonzalocasas @Carter12s @memrecakal . At least regarding the changes to the protocol specification ([ROSBRIDGE_PROTOCOL.md‎](https://github.com/RobotWebTools/rosbridge_suite/pull/1150/changes#diff-23991fba1d884436ff88d9fa435f9a81f65eebaec386465aca4424be56d0bdff)), not the implementation itself.

bjsowa:
Merging since nobody had any objections.

bjsowa:
@mergifyio backport kilted

bjsowa:
@mergifyio backport jazzy

PR #1152, opened 2026-02-19
fix: rejected goal not returning any data.

ikwilnaarhuisman:
**Public API Changes**
None

**Description**
I was working with Roslibjs when I realized that when a goal is rejected it was never made clear through the rosbridge. Now it lets you know instead of becoming stuck in a permanent loop.

`{"op":"action_result","action":"\/topological\/move","values":"Action goal was rejected","status":0,"result":false,"id":"send_action_goal:\/topological\/move:18acdc62-9e38-4968-94f1-b471f226c524"}`|

@YannickdeHoop


bjsowa:
Closing this as it is already fixed by #1139

PR #1154, opened 2026-02-22
fix: Handle action rejection and server timeout (backport #1139)

bjsowa:


bjsowa:
@MatthijsBurgh Is it ok if I bypass review requirements for backports?

MatthijsBurgh (approved):

PR #1157, opened 2026-02-22
feat: Add an option to use EventsExecutor

bjsowa:
**Public API Changes**
Adds `use_events_executor` parameter to Rosbridge server.

**Description**
EventsExecutor is an experimental implementation of ROS executor that use callback functions of ROS entities to push events into the event queue. This significantly reduces the amount of maintenance operations in comparison to classic wait-set approach. In applications where ROS must process many messages from its subscribers, services or actions, using EventsExecutor can greatly reduce the CPU consumption.

This PR adds the `use_events_executor` parameter to the Rosbridge server, which, when enabled, runs the node in EventsExecutor instead of SingleThreadedExecutor. The tests have been adjusted to be run both with this parameter disabled and enabled.

### DIFF
--- a/rosbridge_server/CMakeLists.txt
+++ b/rosbridge_server/CMakeLists.txt
@@ -22,7 +22,7 @@ install(FILES
 
 if(BUILD_TESTING)
   find_package(launch_testing_ament_cmake REQUIRED)
-  
+
   # Run each test with both SingleThreadedExecutor and EventsExecutor
   set(TEST_FILES
     test/websocket/advertise_action.test.py
@@ -35,7 +35,7 @@ if(BUILD_TESTING)
     test/websocket/best_effort_publisher.test.py
     test/websocket/multiple_subscribers_raw.test.py
   )
-  
+
   foreach(TEST_FILE ${TEST_FILES})
     # Extract base name for test naming
     get_filename_component(TEST_NAME ${TEST_FILE} NAME_WE)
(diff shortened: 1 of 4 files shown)


sea-bass (approved):
Awesome -- this is very exciting. How does this further reduce CPU usahe?

I believe the EventsExecutor is on track to be non experimental with the next LTS (Lyrical), and if that happens that should probably be another branching point... or at least switching the default flag to true.

bjsowa:
> Awesome -- this is very exciting. How does this further reduce CPU usahe?

When I tried it last year, it reduced the usage by 50% for me. But this probably depends on the way rosbridge is used. In applications which pass many small messages, I would expect huge improvement due to the time saved on spinning the executor. In applications which send large messages not so frequently, the reduction probably won't be that noticeable as most of the CPU resources is allocated to message conversion.

> I believe the EventsExecutor is on track to be non experimental with the next LTS (Lyrical), and if that happens that should probably be another branching point... or at least switching the default flag to true.

Ah yes! The long-awaited moment when I'll have to support 5 ROS distributions at once.

If the EventsExecutor becomes non-experimental, I'm up for branching out from kilted and changing the default flag to true for Lyrical and Rolling.

sea-bass:
> Ah yes! The long-awaited moment when I'll have to support 5 ROS distributions at once.

This is why they pay you the big bucks!

PR #1160, opened 2026-02-25
fix: max message size not included in tornado settings.

ikwilnaarhuisman:
The max message size used to be implicitly used by the tornado settings, but this was broken in commit https://github.com/RobotWebTools/rosbridge_suite/commit/949d7c3457628b34d117c78738b8fa2f3514b4bd. 

Because what used to happen is that the tornado.web.Application could implicitly find the attribute about the max_message_size, but because this was moved to a dictionary it can no longer. Causing Tornado to automatically disconnect when it receives a message exceeding 10 megabytes.

@yannickdehoop


bjsowa:
It would be helpful to add a test to rosbridge_server which tests this behavior so this does not happen in the future. Could you write one?

ikwilnaarhuisman:
I had a look at implementing a unit test where you could assert the max_message_size with the tornado max_message_size, but because tornado doesn't expose its parameters at runtime. I can't access it. I'm not sure how to unit test it without adding extra functionality.

ikwilnaarhuisman:
Please note that when tornado receives messages bigger than the `max_message_size` the connection is terminated. Making this is a very annoying issue.

https://github.com/tornadoweb/tornado/blob/3dfc965b2828fb2e48b700b1436dfbd16224a05d/tornado/websocket.py#L323

YannickdeHoop:
@bjsowa Do you have any suggestions with regard to testing?

bjsowa:
> @bjsowa Do you have any suggestions with regard to testing?

If we want to define a certain behavior, IMO we should write an integration test which asserts this behavior, for example, by sending large messages to the server and testing how different values for `max_message_size` affect it.

bjsowa:
The test got stuck so I cancelled the workflow

ikwilnaarhuisman:
They run locally (on jazzy) without a problem.

Other than mypy, whatever happens there.

ikwilnaarhuisman:
@bjsowa I removed some floating debug code. Be sure to quit the workflow if required

PR #1161, opened 2026-02-26
fix: numpy.ndarray not handled in CBOR serialization

### LINKED ISSUE #965: Error on CBOR compression of custom messages
**Description**

By using `Roslibjs` to subscribe to topics using `CBOR` compression, I'm having a `Rosbridge` error with some custom messages. I can subscribe to standard messages with the compression, but not some of my custom ones, particularly when containing vectors of other custom messages.

- Library Version: Last commit of `humble` branch.
- ROS Version: `Humble`
- Platform / OS: `Ubuntu 22.04`

**Expected Behavior**

All messages arriving through `Roslibjs` on the wep application, when `CBOR` compression is requested.

**Actual Behavior**

The messages don't arrive, and `Rosbridge` prints the following error:
```
[rosbridge_websocket-4]     for slot, slot_type in msg.get_fields_and_field_types().items():
[rosbridge_websocket-4] AttributeError: 'numpy.ndarray' object has no attribute 'get_fields_and_field_types'
```
I don't know where this `numpy.array` comes from since my nodes are `C++`. I'm guessing is some conversion to `Python` on the message generators in between.

**Possible solution**

I was able to got it to work, but I'm unsure of the solution. Or, if the problem is caused by another completely different issue and this workaround is not the best approach. 

On `cbor_conversion.py`, I added to the beginning of `extract_cbor_values` function:

```python
if isinstance(msg, np.ndarray):
    return msg.tolist() 
```

And imported numpy. The final result being:

```python
# (...)

import numpy as np

# (...)

def extract_cbor_values(msg):
    """Extract a dictionary of CBOR-friendly values from a ROS message.

    Primitive values will be casted to specific Python primitives.

    Typed arrays will be tagged and packed into byte arrays.
    """

    if isinstance(msg, np.ndarray):
        return msg.tolist() 

    out = {}
    for slot, slot_type in msg.get_fields_and_field_types().items():
        val = getattr(msg, slot)

        # string
        if slot_type in STRING_TYPES:
            out[slot] = str(val)

        # bool
        elif slot_type in BOOL_TYPES:
            out[slot] = bool(val)

        # integers
        elif slot_type in INT_TYPES:
            out[slot] = int(val)

        # floats
        elif slot_type in FLOAT_TYPES:
            out[slot] = float(val)

        # time/duration
        elif slot_type in TIME_TYPES:
            out[slot] = {
                "sec": int(val.sec),
                "nanosec": int(val.nanosec),
            }

        # byte array
        elif slot_type in BYTESTREAM_TYPES:
            out[slot] = bytes(val)

        # bool array
        elif slot_type in BOOL_ARRAY_TYPES:
            out[slot] = [bool(i) for i in val]

        elif slot_type in STRING_ARRAY_TYPES:
            out[slot] = [str(i) for i in val]

        # numeric arrays
        elif slot_type in TAGGED_ARRAY_FORMATS:
            tag, fmt = TAGGED_ARRAY_FORMATS[slot_type]
            fmt_to_length = fmt.format(len(val))
            packed = struct.pack(fmt_to_length, *val)
            out[slot] = Tag(tag=tag, value=packed)

        # array of messages
        elif type(val) in LIST_TYPES:
            out[slot] = [extract_cbor_values(i) for i in val]

        # message
        else:
            out[slot] = extract_cbor_values(val)

    return out
```

It seems to be working for every message I request, and my custom messages are being processed much faster with the `compression='cbor'` flag now.

Any tips or suggestions?

Thanks!

atofetti-botbot:
Fix numpy.ndarray not handled in CBOR serialization

Closes [#965](https://github.com/RobotWebTools/rosbridge_suite/issues/965) 

When `cbor_compression` is enabled, ROS 2 stores fixed-size primitive array fields (e.g. `float64[36]` covariance in `Odometry`/`Imu`, `float32[3]` position in `SportModeState`) as `numpy.ndarray` internally. These types are not covered by `TAGGED_ARRAY_FORMATS`, which only handles `sequence<*>` types, and `numpy.ndarray` is not in `LIST_TYPES`. They fall through to the `else` branch which calls `extract_cbor_values(val)` on the array, triggering:

    AttributeError: 'numpy.ndarray' object has no attribute 'get_fields_and_field_types'

The fix adds one `elif` to the existing type dispatch chain, directly after the `TAGGED_ARRAY_FORMATS` handler where it logically belongs. `hasattr(val, 'tolist')` identifies numpy arrays without requiring an import.

Note: `message_conversion.py` already handles this correctly — its `list_types` includes `np.ndarray`. This PR brings `cbor_conversion.py` into parity.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/cbor_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/cbor_conversion.py
@@ -99,7 +99,7 @@ def extract_cbor_values(msg: ROSMessage) -> dict[str, Any]:
             out[slot] = CBORTag(tag=tag, value=packed)
 
         # fixed-size primitive arrays
-        elif hasattr(val, 'tolist'):
+        elif hasattr(val, "tolist"):
             out[slot] = val.tolist()
 
         # array of messages
--- a/rosbridge_library/src/rosbridge_library/internal/cbor_conversion.py
+++ b/rosbridge_library/src/rosbridge_library/internal/cbor_conversion.py
@@ -98,6 +98,10 @@ def extract_cbor_values(msg: ROSMessage) -> dict[str, Any]:
             packed = struct.pack(fmt_to_length, *val)
             out[slot] = CBORTag(tag=tag, value=packed)
 
+        # fixed-size primitive arrays
+        elif hasattr(val, 'tolist'):
+            out[slot] = val.tolist()
+
         # array of messages
         elif type(val) in LIST_TYPES:
             out[slot] = [extract_cbor_values(i) for i in val]
--- a/rosbridge_library/test/internal/test_cbor_conversion.py
+++ b/rosbridge_library/test/internal/test_cbor_conversion.py
@@ -205,7 +205,6 @@ class TestCBORConversion(unittest.TestCase):
             self.assertEqual(type(key), str)
 
     def test_numpy_array(self) -> None:
-
         class FakeMsg:
             def get_fields_and_field_types(self) -> dict[str, str]:
                 return {"data": "float64[3]"}
--- a/rosbridge_library/test/internal/test_cbor_conversion.py
+++ b/rosbridge_library/test/internal/test_cbor_conversion.py
@@ -5,6 +5,7 @@ import struct
 import unittest
 from array import array
 
+import numpy as np
 from builtin_interfaces.msg import Duration, Time
 from cbor2 import CBORTag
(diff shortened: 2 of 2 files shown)


bjsowa:
Care to add a check for this to `test_cbor_conversion.py` ?

bjsowa (approved):
lgtm. The mypy errors in Rolling distribution are not related to this PR

PR #1164, opened 2026-02-26
fix: Use new interface base classes for type checking

bjsowa:
ros2/rosidl#887, ros2/rosidl#912, ros2/rosidl_python#230 and ros2/rosidl_python#241 made it so that all generated interfaces derive from common abstract base classes (finally proper polymorphism). This means we can drop the usage of Protocol classes for type checking. 
 
 The only problem is, the changes are not backported to Kilted (and probably won't be). I tried to make it compatible between distributions but that only made it look ugly and was not working properly. This means we probably need to branch out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,7 @@ jobs:
         include:
           # Test supported ROS 2 distributions
           # https://docs.ros.org/en/rolling/Releases.html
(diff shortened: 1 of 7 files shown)


bjsowa:
@sea-bass @MatthijsBurgh Are we ok with branching out for Kilted?

bjsowa:
Related: #1166

sea-bass (approved):
Fine if you would like to branch again! Anyway Kilted will be EOL later this year.

PR #1172, opened 2026-03-09
feat: Drop support for BSON

bjsowa:
**Public API Changes**
Removed `binary_encoder_type` and `bson_only_mode` parameters.

**Description**
As discussed in RobotWebTools/roslibjs#1039, #1105 and #1124, there are many reasons for dropping support for BSON and once we figure a way for the server to be able to receive CBOR encoded messages, not only send them, there will be no real advantage of using BSON over CBOR.

Since Lyrical is right around the corner, I figured this is the best time for feature-breaking PRs. This WILL NOT be backported to Kilted or earlier distros.

### DIFF
--- a/rosbridge_library/package.xml
+++ b/rosbridge_library/package.xml
@@ -19,7 +19,6 @@ ROS action, like subscribe, publish, call service, and interact with params.
   <buildtool_depend>ament_cmake_ros</buildtool_depend>
   <buildtool_depend>ament_cmake_python</buildtool_depend>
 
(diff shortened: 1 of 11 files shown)


EzraBrooks (approved):
Agree. Should also do this in roslibjs 3.x (recently referred to in RobotWebTools/roslibjs#1171)

bjsowa:
@EzraBrooks Do you think we should also drop PNG support?

PR #1173, opened 2026-03-10
fix: Deadlock in concurrent module import in ros_loader

FieldSwan:
**Public API Changes**
None

**Description**
The new version of rosbridge, `3.1.0` introduced a regression thanks to the new execution approach. Clients that connect too quickly can cause the bridge to get stuck in an intermediate state, requiring a node restart to fix. This fixes that by moving around the server initialization order and adding a lock to the interface imports. Tests are included to compare results with and without this fix.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/ros_loader.py
+++ b/rosbridge_library/src/rosbridge_library/internal/ros_loader.py
@@ -57,6 +57,7 @@ _loaded_actions: dict[str, type[ROSAction]] = {}
 _msgs_lock = Lock()
 _srvs_lock = Lock()
 _actions_lock = Lock()
+_import_lock = Lock()
 
 
 class InvalidTypeStringException(Exception):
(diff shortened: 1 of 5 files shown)


bjsowa:
Thank you for your contribution! Haven't looked through the whole code but seems like a neatly written fix. 

Just one thing: all contributions should be made on top of `ros2` branch (which targets only Rolling at the moment) and once they merged I backport them to other distros. Could you rebase your commit and change the base branch to `ros2`?

FieldSwan:
> Thank you for your contribution! Haven't looked through the whole code but seems like a neatly written fix.
> 
> Just one thing: all contributions should be made on top of `ros2` branch (which targets only Rolling at the moment) and once they merged I backport them to other distros. Could you rebase your commit and change the base branch to `ros2`?

Thanks! Ran into this issue inconsistently on our project and finally tracked down the cause.

I'll go ahead and rebase; been fighting the linter, but should be done soon.

FieldSwan:
Rebased to ros2 branch. Should be OK now.

bjsowa:
The CI workflow did not trigger after changing the base branch. Could you try force pushing the commits again or pushing an empty commit to trigger it?

FieldSwan:
GH actions checks passed. Let me know if you need anything else!

bjsowa (approved):
LGTM. The tests are a lot more than I would expect for such a simple fix but certainly a welcome surprise. The `startup_race` test is passing for me even without the fix but let's leave it in as it might detect some other regressions in the future.

Just out of curiosity, did you use Generative AI to write the tests?

FieldSwan:
> LGTM. The tests are a lot more than I would expect for such a simple fix but certainly a welcome surprise. The `startup_race` test is passing for me even without the fix but let's leave it in as it might detect some other regressions in the future.
> 
> Just out of curiosity, did you use Generative AI to write the tests?

Yeah, used Claude Opus 4.6 and then made edits to reduce some of the tests and verbosity. It initially produced some useless tests, but this seemed correct/useful to me; I was careful to check and test everything. I did see the single threaded startup race test (but not event executor test) and concurrent import test failing on kilted 3.1.0 but working with the fix in-place but maybe there is some machine dependency. 

I also tested the actual issue replication (have client connecting ASAP on node startup on a slower machine) with and without the fix and saw it work; I think the server creation order side of the fix is more important than the import lock.

PR #1179, opened 2026-03-11
fix: Deadlock in concurrent module import in ros_loader (backport #1173)

bjsowa:
backport of #1173


bjsowa:
@mergifyio backport humble

PR #1184, opened 2026-03-16
Configurable qos_depth

FieldSwan:
**Public API Changes**
None

**Description**
Adds an option to define default bridge subscriber QoS depth at launch time. This was added because in our usecase we do not care about any old data and preferred a queue size of 1.


bjsowa:
I'm holding up any PRs related to QoS till #1150 gets merged

Issue #1144, opened 2026-01-29
Failed to add 'rcl_action_client_t' to wait set: action client pointer is invalid


szobov:
**Description**

- Library Version:
```
commit 5204927d20d39fe5310c6a316cd6d26a2912d928 (HEAD -> ros2, origin/ros2, origin/HEAD)
Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
```
- ROS Version: jazzy
- Platform / OS: ubuntu 24.04

**Steps To Reproduce**
We run rosbridge as a separate node and for some of the actions calls we're getting:
```
[rosbridge_websocket-11] Exception in callback <function main.<locals>.spin_ros at 0x7d8d7a393ec0>
[rosbridge_websocket-11] Traceback (most recent call last):
[rosbridge_websocket-11]   File "/root/chimera/.env/lib/python3.12/site-packages/tornado/ioloop.py", line 937, in _run
[rosbridge_websocket-11]     val = self.callback()
[rosbridge_websocket-11]           ^^^^^^^^^^^^^^^
[rosbridge_websocket-11]   File "/root/chimera/docker_workspace/install/rosbridge_server/lib/rosbridge_server/rosbridge_websocket", line 244, in spin_ros
[rosbridge_websocket-11]     executor.spin_once(timeout_sec=0.01)
[rosbridge_websocket-11]   File "/root/chimera/docker_workspace/install/rclpy/lib/python3.12/site-packages/rclpy/executors.py", line 923, in spin_once
[rosbridge_websocket-11]     self._spin_once_impl(timeout_sec)
[rosbridge_websocket-11]   File "/root/chimera/docker_workspace/install/rclpy/lib/python3.12/site-packages/rclpy/executors.py", line 904, in _spin_once_impl
[rosbridge_websocket-11]     handler, entity, node = self.wait_for_ready_callbacks(
[rosbridge_websocket-11]                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rosbridge_websocket-11]   File "/root/chimera/docker_workspace/install/rclpy/lib/python3.12/site-packages/rclpy/executors.py", line 874, in wait_for_ready_callbacks
[rosbridge_websocket-11]     return next(self._cb_iter)
[rosbridge_websocket-11]            ^^^^^^^^^^^^^^^^^^^
[...]

PR #1183, opened 2026-03-16
fix: Prevent client destruction race condition

FieldSwan:
**Public API Changes**
None

**Description**
1. Adds a client creation/destruction stress test to replicate the issue observed where sometimes we would see:
```bash
Exception in thread Thread-1 (spin):
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/executors.py", line 374, in spin
    self.spin_once()
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/executors.py", line 968, in spin_once
    self._spin_once_impl(timeout_sec)
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/executors.py", line 951, in _spin_once_impl
    handler, entity, node = self.wait_for_ready_callbacks(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/executors.py", line 921, in wait_for_ready_callbacks
    return next(self._cb_iter)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/executors.py", line 820, in _wait_for_ready_callbacks
    waitable.add_to_wait_set(wait_set)
  File "/opt/ros/kilted/lib/python3.12/site-packages/rclpy/event_handler.py", line 176, in add_to_wait_set
rclpy._rclpy_pybind11.InvalidHandle: cannot use Destroyable because destruction was requested
```
Note that this stress test is relatively slow, takes 20 seconds on my machine.

2. Fixes the issue by adding the websocket protocol destruction to the executor waitset (if possible) to prevent a race condition with the executor's other tasks.

Tested on ROS2 kilted.

May be related to: https://github.com/RobotWebTools/rosbridge_suite/issues/1144

PR #1185, opened 2026-03-16
Fix race condition in websocket test

FieldSwan:
**Public API Changes**
None

**Description**
This fixes an issue with the websocket testing instrumentation which had a race condition inherent in it that was causing my test to sometimes fail incorrectly (false negative) which led me to believe that other solutions were helping with other issues (they were not).

Changes:
1. Fix websocket testing instrumentation race condition when many clients requested at once. You can verify this fix by trying this before and after the fix: 
```bash
colcon test --packages-select rosbridge_server --ctest-args -R "test_startup_race" --repeat until-fail:500 --event-handlers console_direct+ 2>&1 | tee /tmp/startup_race_retry.txt
# in another terminal:
stress --cpu 15
```
Eventually this should fail, especially with stress running at the same time (load your CPU as much as possible). The fix was to use the reactorThread for sending messages and to add retries to getting the ros port parameter from the ros node.
3. Removes the `startup_race.test.py` test which, it turns out, was mostly exercising the websocket test instrumentation and not any particularly useful code paths. 
4. Reverts the previous "fix" I made to the server instantiation order. Upon further examination, I realized that the order does not matter since the websocket requests are not processed in `rosbridge_websocket.py` until we call `await stop_event.wait()`, which is already after the ros2 executor and node are all started. I can remove this change if we want to keep the order of instantiation I had changed to earlier (either way should work fine).


bjsowa:
You sure you want to remove the `startup_race` test? Didn't it help you detect the race condition you are trying to fix now?

ikwilnaarhuisman:
Tested this code, doesn't fix the underlying issue mentioned in. https://github.com/RobotWebTools/rosbridge_suite/issues/1144

Happens because the RCLPY (SingleThreaded)Executor may crash, which then still throws `[rosbridge_websocket-11] rclpy._rclpy_pybind11.RCLError: Failed to add 'rcl_action_client_t' to wait set: action client pointer is invalid, at ./src/rcl_action/action_client.c:510`

Me and @YannickdeHoop are working on an actual fix which uses the executor directly so you may never have race conditions for your results ever again: https://github.com/eurogroep/rosbridge_suite/tree/fix/use-executor-callback-queue

@bjsowa LMK what you think of this.

bjsowa:
> @bjsowa LMK what you think of this.

A test, demonstrating where the underlying issue is, would be helpful.

FieldSwan:
> You sure you want to remove the `startup_race` test? Didn't it help you detect the race condition you are trying to fix now?

It did help me identify a race condition, but specifically the race condition located in `rosbridge_server/test/websocket/common.py`, not in the functional code. 

The test does not capture the original deadlock condition I found; `rosbridge_library/test/internal/test_ros_loader_concurrent.py` does though. There's technically no harm in leaving the test in so I could go either way on this, just wanted it to be clear that it's more similar to a smoke test that happened to catch the test instrumentation randomly than a functional test that is deterministically finding future issues.

bjsowa (approved):

FieldSwan:
> Tested this code, doesn't fix the underlying issue mentioned in. #1144
> 
> Happens because the RCLPY (SingleThreaded)Executor may crash, which then still throws `[rosbridge_websocket-11] rclpy._rclpy_pybind11.RCLError: Failed to add 'rcl_action_client_t' to wait set: action client pointer is invalid, at ./src/rcl_action/action_client.c:510`
> 
> Me and @YannickdeHoop are working on an actual fix which uses the executor directly so you may never have race conditions for your results ever again: https://github.com/eurogroep/rosbridge_suite/tree/fix/use-executor-callback-queue
> 
> @bjsowa LMK what you think of this.

@ikwilnaarhuisman did you try PR https://github.com/RobotWebTools/rosbridge_suite/pull/1183 ? This is the fix I suggested could fix Issue #1144

bjsowa:
@Mergifyio backport kilted

bjsowa:
@mergifyio backport jazzy

bjsowa:
@mergifyio backport humble

PR #1201, opened 2026-03-27
feat: Cache rosapi get/set parameter clients

bjsowa:
**Public API Changes**
None

**Description**
Retrieving or setting multiple parameters from/to a node using rosapi right now spawn a client for each parameter retrieved/set. Spawning multiple clients and destroying them after each call does not scale well, especially with DDS which can flood the network with discovery messages.

This PR implements a caching mechanism for get/set service clients in rosapi. Setting multiple parameters for the same node will reuse the same client. The clients are automatically cleaned up after 5 seconds of inactivity.

### DIFF
--- a/rosapi/src/rosapi/params.py
+++ b/rosapi/src/rosapi/params.py
@@ -35,7 +35,7 @@ from __future__ import annotations
 import fnmatch
 from dataclasses import dataclass
 from json import dumps, loads
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, cast
 
 from rcl_interfaces.msg import Parameter, ParameterType, ParameterValue
 from rcl_interfaces.srv import GetParameters, ListParameters, SetParameters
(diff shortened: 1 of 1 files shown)


rosapi/src/rosapi/params.py
```diff
@@ -81,7 +86,21 @@
 ]
 
 
-def init(node: Node, timeout_sec: float = DEFAULT_PARAM_TIMEOUT_SEC) -> None:
+@dataclass
+class _CachedClient:
+    use_count: int
+    last_used_time: Time
+    client: Client
```
sea-bass, line 104: docstrings would be helpful here

rosapi/src/rosapi/params.py
```diff
@@ -91,16 +110,80 @@ def init(node: Node, timeout_sec: float = DEFAULT_PARAM_TIMEOUT_SEC) -> None:
...
+    """
+    Get a cached client for the given service, or create a new one if it doesn't exist.
+
+    :param service_name: The name of the service to get a client for.
+    :param service_type: The type of the service to get a client for.
+    :return: A client for the given service.
+    :rtype: Client
+    """
+    if service_name in _cached_clients:
+        return _cached_clients[service_name].client
```
sea-bass: nit: this leads to 2 dictionary lookups. Consider doing e.g.

```python
if (cached_client := _cached_clients.get(service_name) is not None):
    return cached_client.client
```

or if you don't want to use walrus operator make the intermediate variable, as you wish

sea-bass (approved):
nice! minor optional comments as usual

rosapi/src/rosapi/params.py
bjsowa: To be honest, I forgot this operator exists. In all fairness, I prefer the code to be as much accessible to other users as I can, so I'll go with the intermediate variable.

bjsowa:
@mergifyio backport kilted jazzy

PR #1203, opened 2026-03-27
feat: Create topic registrations in publish when topic not previously advertised

bjsowa:
**Public API Changes**
The `publish` message now respects `type` and `id` fields

**Description**
Since `publish` already registers a topic even with no previous call to `advertise`, I figured it would make sense to unify the tracking of topic registration between these two capabilities. `publish` capability now basically becomes an extension of `advertise` capability, supporting the same fields with the addition of `msg`.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/capabilities/advertise.py
+++ b/rosbridge_library/src/rosbridge_library/capabilities/advertise.py
@@ -67,7 +67,11 @@ class Registration:
         manager.unregister(self.client_id, self.topic)
 
     def register_advertisement(
-        self, msg_type: str, adv_id: str | None = None, latch: bool = False, queue_size: int = 100
+        self,
+        msg_type: str | None,
+        adv_id: str | None = None,
+        latch: bool = False,
+        queue_size: int = 100,
     ) -> None:
         # Register with the publisher manager, propagating any exception
         manager.register(
@@ -111,8 +115,6 @@ class Advertise(Capability):
         protocol.register_operation("advertise", self.advertise)
         protocol.register_operation("unadvertise", self.unadvertise)
 
-        self._registrations: dict[str, Registration] = {}
-
     def advertise(self, message: dict[str, Any]) -> None:
         # Pull out the ID
         aid = message.get("id")
@@ -144,12 +146,16 @@ class Advertise(Capability):
             self.protocol.log("debug", "No topic security glob, not checking advertisement.")
 
         # Create the Registration if one doesn't yet exist
-        if topic not in self._registrations:
+        if topic not in self.protocol.topic_registrations:
             client_id = self.protocol.client_id
-            self._registrations[topic] = Registration(client_id, topic, self.protocol.node_handle)
-
-        # Register, propagating any exceptions
-        self._registrations[topic].register_advertisement(msg_type, aid, latch, queue_size)
+            registration = Registration(client_id, topic, self.protocol.node_handle)
+            registration.register_advertisement(msg_type, aid, latch, queue_size)
+            self.protocol.topic_registrations[topic] = registration
+        else:
+            # Register, propagating any exceptions
+            self.protocol.topic_registrations[topic].register_advertisement(
+                msg_type, aid, latch, queue_size
+            )
 
     def unadvertise(self, message: dict[str, Any]) -> None:
         # Pull out the ID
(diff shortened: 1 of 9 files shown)


rosbridge_library/src/rosbridge_library/capabilities/advertise.py
```diff
@@ -144,12 +146,16 @@ def advertise(self, message: dict[str, Any]) -> None:
...
         # Create the Registration if one doesn't yet exist
-        if topic not in self._registrations:
+        if topic not in self.protocol.topic_registrations:
             client_id = self.protocol.client_id
-            self._registrations[topic] = Registration(client_id, topic, self.protocol.node_handle)
-
-        # Register, propagating any exceptions
-        self._registrations[topic].register_advertisement(msg_type, aid, latch, queue_size)
+            registration = Registration(client_id, topic, self.protocol.node_handle)
+            registration.register_advertisement(msg_type, aid, latch, queue_size)
```
sea-bass: I feel like if you remove this line, the `else` condition can be de-indented again and just run in every case. It would be equivalent, closer to the old code, and less repetitive, right?

rosbridge_library/src/rosbridge_library/capabilities/publish.py
```diff
@@ -57,13 +64,12 @@ def __init__(self, protocol: Protocol) -> None:
         # Register the operations that this capability provides
         protocol.register_operation("publish", self.publish)
 
-        # Save the topics that are published on for the purposes of unregistering
-        self._published: dict[str, bool] = {}
-
     def publish(self, message: dict[str, Any]) -> None:
         # Do basic type checking
         self.basic_type_check(message, self.publish_msg_fields)
+        aid: str | None = message.get("id")
```
sea-bass: can we just call this `id` or `msg_id`? It seems the `aid` naming came from the advertising code and is kind of confusing, but in this file it should for sure not have an `a` in front of it for clarity

rosbridge_library/src/rosbridge_library/capabilities/publish.py
```diff
@@ -86,32 +92,20 @@ def publish(self, message: dict[str, Any]) -> None:
...
-            self.protocol.node_handle,
-            latch=latch,
-            queue_size=queue_size,
-        )
-
-    def finish(self) -> None:
-        client_id = self.protocol.client_id
-        for topic in self._published:
-            manager.unregister(client_id, topic)
-        self._published.clear()
```
sea-bass, line 117: so this can be removed because now the `advertise`'s `finish()` function clears it since it's all available in the protocol, right?

rosbridge_library/src/rosbridge_library/internal/topics.py
```diff
@@ -43,6 +43,14 @@ def __init__(self, topic: str) -> None:
         )
 
 
+class TopicNotRegisteredException(Exception):
+    def __init__(self, topic: str) -> None:
+        Exception.__init__(
+            self,
+            f"Cannot publish to topic {topic} as it is not registered",
```
sea-bass: nit: add single quotes around the topic name for readability?

rosbridge_server/test/websocket/common.py
```diff
@@ -133,14 +133,20 @@ def run_websocket_test(
     executor.add_node(node)
 
     async def task() -> None:
-        await test_fn(node, lambda: connect_to_server(node))
-        reactor.callFromThread(reactor.stop)  # type: ignore[attr-defined]
+        try:
+            await test_fn(node, lambda: connect_to_server(node))
+        finally:
+            reactor.callFromThread(reactor.stop)  # type: ignore[attr-defined]
```
sea-bass, line 139: why did the `try` need to be added here? is it supposed to raise in some test cases? If so, worth adding some explanation

sea-bass (commented):
Looks good! Left some minor suggestions.

rosbridge_library/src/rosbridge_library/capabilities/advertise.py
bjsowa: The problem I'm trying to solve here is that if `register_advertisement` throws an exception (e.g. wrong message type) we would still put the registration into `topic_registrations`, even when no advertisement was made.
The `publish` capability checks whether there is a topic registration to determine whether it needs to make one itself.
This change defers putting the registration into `topic_registrations` until there is at least one registered advertisement.

rosbridge_library/src/rosbridge_library/capabilities/publish.py
bjsowa: `id` would interfere with Python's builtin. `aid` is meant to mean `advertisement ID` and it refers to the same thing in `publish` and `advertise` capabilities. I could change it to `adv_id` or `advertisement_id` for more clarity.

rosbridge_library/src/rosbridge_library/capabilities/publish.py
bjsowa: Yes, `advertise` capability takes care of cleaning all of the topic advertisements

rosbridge_library/src/rosbridge_library/internal/topics.py
bjsowa: Should I add this to other Exceptions as well?

rosbridge_server/test/websocket/common.py
bjsowa: If test throws an exception (for example, assertEqual fails), `reaction.stop` is never called and the test hangs until a timeout.

rosbridge_library/src/rosbridge_library/capabilities/publish.py
sea-bass: Ah, I see! Maybe one of those names would help then

rosbridge_library/src/rosbridge_library/internal/topics.py
sea-bass: Nah, don't worry, it's fine to just change this one

sea-bass (approved):

bjsowa:
@mergifyio backport kilted

bjsowa:
@mergifyio backport jazzy

bjsowa:
@mergifyio backport humble

PR #1220, opened 2026-04-05
fix: Auto-populate header.stamp when header omitted

bjsowa:
**Public API Changes**
None

**Description**
rosbridge already fills the `header.stamp` field with a current time when `stamp` is omitted but it doesn't work in situation when client omits the entire `header`. This PR fixes this issue. It also clarifies the usage of `"now"` string in stamp fields and adds tests that check the behavior.


sea-bass (approved):
LGTM!

bjsowa:

PR #1228, opened 2026-05-05
fix: Don't use MultiThreadedExecutor in tests

bjsowa:
**Public API Changes**
None

**Description**
Action capability test seems to be flaky again. Changing from MultiThreadedExecutor to SingleThreadedExecutor seems to fix that issue so I replaced it whenever I could find it.

### DIFF
--- a/rosbridge_library/test/capabilities/test_action_capabilities.py
+++ b/rosbridge_library/test/capabilities/test_action_capabilities.py
@@ -10,7 +10,7 @@ from typing import Any
 import rclpy
 from action_msgs.msg import GoalStatus
 from example_interfaces.action._fibonacci import Fibonacci_FeedbackMessage
-from rclpy.executors import MultiThreadedExecutor
+from rclpy.executors import SingleThreadedExecutor
 from rclpy.node import Node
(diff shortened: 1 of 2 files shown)


bjsowa:
CI seems to be broken for now due to Rolling switching from Ubuntu 24 to Ubuntu 26

bjsowa:
Test failures are expected for now, this only fixes one of them

sea-bass (approved):
At some point in the past, multithreaded executor was needed for the tests to even pass. But quite possibly the backend has improved substantially where this is no longer needed. So if CI passes, good!

Edit: Just saw your comment. So I think ROS 2 actions need a multithreaded executor to process cancel callbacks while an action is executing. If you can find a way around that, that would probably be the solution.

bjsowa:
> At some point in the past, multithreaded executor was needed for the tests to even pass. But quite possibly the backend has improved substantially where this is no longer needed. So if CI passes, good!
> 
> Edit: Just saw your comment. So I think ROS 2 actions need a multithreaded executor to process cancel callbacks while an action is executing. If you can find a way around that, that would probably be the solution.

All the tests related to actions seem to pass so I assume multi threaded executor is no longer needed. We'll see if this can be backported.

bjsowa:

PR #1230, opened 2026-05-07
chore: Use resolute container and ros2-testing for Rolling, add Lyrical

bjsowa:
**Public API Changes**
None

**Description**
Rolling recently switched to Ubuntu Resolute and Lyrical was made available as well. Since Github does not provide ubuntu-26.04 runner yet, we have to use a container. Also, Rolling and Lyrical packages are not yet available on ros2 package repository so we have to use ros2-testing repo.

Failing tests are not related to CI and will be fixed in another PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         with:
           path: ros_ws/src
 
-      - uses: ros-tooling/setup-ros@v0.7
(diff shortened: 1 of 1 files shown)


sea-bass (approved):
I'd personally just wait until Ubuntu 26.04 containers are available on GitHub, but that's an easy change to make later. LGTM!

PR #1238, opened 2026-05-08
fix: replace 4 bare except clauses with except Exception

KeloYuan:
Replaces 4 bare `except:` with `except Exception:` in 3 files: test_subscription_modifiers.py (2), test_services.py (1), websocket_handler.py (1).


bjsowa (changes requested):
Remove the `noqa` comment to fix lint error

PR #1247, opened 2026-05-14
fix: Don't raise exception in service callback

bjsowa:
**Public API Changes**
None

**Description**
Since recent rclpy changes, raising an exception in a ROS entity callback crashes the executor. This PR changes how shutdown of services work during a live request.

PS. This fixes one test failure. Still a couple left to fix


sea-bass (approved):
As long as this doesn't cause the service to actually respond with a default, it's okay!

bjsowa:
@mergifyio backport kilted

bjsowa:
@mergifyio backport jazzy humble

PR #1248, opened 2026-05-14
feat: Improve action unadvertising

bjsowa:
**Public API Changes**
None

**Description**
Previously, action servers were never actually destroyed and just left hanging. This PR improves the logic so that now, when an action gets unadvertised:
1. Any new goals are rejected.
2. Pending goals are aborted
3. When the last goal finishes, action server destruction is scheduled 1 second into the future to allow all action results to be sent.

The test confirms that the user can advertise the action, unadvertise the action and then advertise the action with the same name again.

I'm still not sure whether `action_server.destroy()` is perfectly safe but I didn't experience any crashes when testing it.


sea-bass (approved):

bjsowa:

PR #1256, opened 2026-05-18
fix: Prevent client destruction race in services.call_service (backport #1255)

JWhitleyWork:
Back-port of #1255 to `jazzy`. `rosbridge_library/internal/services.py` is identical between `ros2` and `jazzy`, so the patch cherry-picks cleanly with no conflicts and the diff is the same. See #1255 for the full rationale, root cause analysis, and regression-test description.

In short: `services.call_service` creates and destroys an `rclpy.Client` directly from a worker thread, racing the executor that is spinning the same node. The race occasionally leaves the executor referencing a freed PyCapsule, and the next rclpy binding call on the node surfaces as:

```
TypeError: Object of type 'NoneType' is not an instance of 'capsule'
```

after which every subsequent `call_service` on the rosbridge process fails the same way until the process is restarted. The fix routes the create/destroy calls through `executor.create_task(...)`, mirroring the same pattern that #1183 applied to the disconnect path.

A new `test/internal/services/test_stress_service_clients.py` regression test exercises the race window (parallel callers, barrier-synchronized start, rapid sequential cycles, interleaved waves) and asserts the executor stays healthy and responsive afterwards.

### DIFF
--- a/rosbridge_library/src/rosbridge_library/internal/services.py
+++ b/rosbridge_library/src/rosbridge_library/internal/services.py
@@ -32,7 +32,7 @@
 from __future__ import annotations
 
 from threading import Event, Thread
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, TypeVar
 
 from rclpy.callback_groups import ReentrantCallbackGroup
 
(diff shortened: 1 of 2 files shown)


bjsowa:
You don't have to create backports yourself. I run mergify to create backports after merging each PR to `ros2` branch, when applicable.

JWhitleyWork:
> You don't have to create backports yourself. I run mergify to create backports after merging each PR to `ros2` branch, when applicable.

I can close these if you want your bot to do it instead. Claude did them for me so it was negligible effort.

JWhitleyWork:
Closing because #1255 got some updates for a related fix.

PR #1290, opened 2026-06-30
fix: Prevent event loop starvation by adding a write queue

### LINKED ISSUE #1243: Version 2.0.6 is significantly slower
**Description**
I've been using `rosbridge_suite` as the bridge between my ROS2 stack and a web UI for several years. Recently, after updating to the lastest `humble`, I noticed a slowdown in the performance.

I traced the issue to version **2.0.6**. It introduces a severe performance regression. However, release 2.0.5 works perfectly fine.

It seems the regression was introduced by commit [d858f8d](https://github.com/RobotWebTools/rosbridge_suite/commit/d858f8d3adb1ef7fb50e7fa806ad317929832037) "fix: Reduce idle CPU consumption of websocket server".

Reverting d858f8d on top of 2.0.6 restores the previous behavior.

- Library Version: 2.0.6
- ROS Version: ROS 2 Humble
- Platform / OS: Ubuntu 22.04.5 LTS, Python 3.10.12

**Expected Behavior**
Same low latency and throughput as in version 2.0.5.

**Actual Behavior**
Starting from 2.0.6, WebSocket message delivery becomes extremely slow. Reverting commit `d858f8d` immediately resolves the issue.

bjsowa:
**Public API Changes**
Added `incoming_queue_size` and `write_queue_size` ROS parameters

**Description**
Fixes #1243

Previously, each rosbridge message send operation was scheduled to the asyncio event loop by a call to `call_coroutine_threadsafe`. When a large number of messages was sent, the event loop was quickly overfilled as the next tasks spawned faster than they could be processed and the number of tasks increased indefinitely.

This caused other client to be unable to open connection or receive any messages.

This has been fixed by adding a write queue and a task which drains this queue. When a new message is to be sent, it is scheduled to be put into the queue (by call to `call_coroutine_threadsafe`) ONLY if the queue is not full, otherwise it is dropped. The write task yields the execution to other tasks by call to `await` for each message, so the other clients don't starve, even when the server is unable to process all messages in time.

The incoming queue was also limited in size to prevent the possibility of memory exhaustion.

The starvation test fails on EventExecutor but for a different reason which I won't cover in this PR.

### DIFF
--- a/rosbridge_server/CMakeLists.txt
+++ b/rosbridge_server/CMakeLists.txt
@@ -30,6 +30,10 @@ if(BUILD_TESTING)
     TIMEOUT 60
   )
 
+  # This test is currently failing with EventsExecutor, so we run it only with SingleThreadedExecutor for now.
+  # TODO(bjsowa): Make this test pass with EventsExecutor.
+  add_launch_test(test/websocket/event_loop_starvation.test.py)
+
   # Run each test with both SingleThreadedExecutor and EventsExecutor
   set(TEST_FILES
     test/websocket/advertise_action.test.py
--- a/rosbridge_server/launch/rosbridge_websocket_launch.xml
+++ b/rosbridge_server/launch/rosbridge_websocket_launch.xml
@@ -16,6 +16,8 @@
   <arg name="max_message_size" default="10000000" />
   <arg name="unregister_timeout" default="10.0" />
 
+  <arg name="incoming_queue_size" default="1000" />
+  <arg name="write_queue_size" default="1000" />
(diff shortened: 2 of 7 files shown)


bjsowa:
Any review, please? @mjimenezm00 @ozanm42 @sea-bass @EzraBrooks Anyone?

ozanm42 (approved):

sea-bass (approved):
LGTM, and seeing that this was tested by a user that encountered the problem.

bjsowa:
@mergifyio backport kilted

PR #1307, opened 2026-08-14
fix(rosapi): Resolve action types without ros2cli

sacovo:
**Public API Changes**

No

**Description**

 /rosapi/action_type has never worked since it was added in #1046 (2.4.0). ros2action.api.get_action_names_and_types() is just node.get_action_names_and_types(), which only exists on ros2cli's DirectNode — with rosapi's plain rclpy node it raises AttributeError, so the service never responds and rosapi_node's bare rclpy.spin() dies with it, taking all of rosapi down.

rclpy.action.get_action_names_and_types() queries the graph directly. Adds a regression test and drops the now-unused ros2action dependency. Please backport.

### DIFF
--- a/rosapi/CMakeLists.txt
+++ b/rosapi/CMakeLists.txt
@@ -24,6 +24,7 @@ install(
 
 if(BUILD_TESTING)
   find_package(ament_cmake_pytest REQUIRED)
+  ament_add_pytest_test(${PROJECT_NAME}_test_action_type test/test_action_type.py)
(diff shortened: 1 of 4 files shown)


bjsowa (approved):
Good idea. I'm all for replacing ros2cli dependencies with plain rclpy which has a stable API.

bjsowa:
@mergifyio backport kilted

bjsowa:
@mergifyio backport jazzy

bjsowa:
@mergifyio backport humble

PR #1308, opened 2026-08-14
fix(rosapi): Keep the param client cache from crashing the node

sacovo:
Clients were cached with a default Time() (SYSTEM_TIME), so the cleanup timer's subtraction against the node clock (ROS_TIME) raised TypeError every 0.5s, and the unavailable-service path destroyed a client while leaving it in the cache.

**Public API Changes**
None

**Description**
Since the param client cache was added in #1201, rosapi crashes shortly after the first /rosapi/get_param or /rosapi/set_param call. Clients are cached with a default Time() (SYSTEM_TIME), so the 0.5 s cleanup timer's subtraction against the node clock (ROS_TIME) raises TypeError: Can't subtract times with different clock types, which propagates out of rclpy.spin() and kills rosapi_node. A fresh client is never ready right after create_client(), so the first call takes the "service is not available" path — which destroys the client but leaves it in the cache, both arming the timer crash and handing back a destroyed client (InvalidHandle) on the next call.

Stamps cached clients from the node clock and drops the cache entry when the client is destroyed. Adds regression tests. Please backport.

### DIFF
--- a/rosapi/CMakeLists.txt
+++ b/rosapi/CMakeLists.txt
@@ -24,6 +24,7 @@ install(
 
 if(BUILD_TESTING)
   find_package(ament_cmake_pytest REQUIRED)
+  ament_add_pytest_test(${PROJECT_NAME}_test_params test/test_params.py)
   ament_add_pytest_test(${PROJECT_NAME}_test_stringify_field_types test/test_stringify_field_types.py)
   ament_add_pytest_test(${PROJECT_NAME}_test_typedefs test/test_typedefs.py)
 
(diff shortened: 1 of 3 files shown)


bjsowa:
That's weird. I never had an issue with that. What you're suggesting makes sense though. (we should always avoid mixing SYSTEM_TIME and ROS_TIME).

I wonder whether a better solution won't be to just use `time.monotonic()` for measuring elapsed time. That way, the mechanism won't be sensitive to system clock jumps. @sacovo What do you think?

sacovo:
`time.monotonic()` makes sense. We experienced the crash under a very specific conditions: calling set/get on a node where the parameter service is not ready. This leaves a stale entry in the cache, and the next request hits the error.

bjsowa (approved):

bjsowa:
@mergifyio backport kilted jazzy

